PR #11, opened 2026-01-17
Fix: Use StashId for Tag Exclusion

sampulsar:
#### Database Migration
NO

#### Description
Remove the year as it serves no purpose. (Field currently left can be removed via a data mitigation but needs to be manual SQL)
Change the title to show more information. 
The StashId for a Tag is now used for exclusions. The contains on a string may not have worked correctly for all tags, and the use of the Id is constant with other exclusions. 

image

#### Todos
- [x] Tests

* Fixes #XXXX

### DIFF
--- a/frontend/src/Settings/ImportLists/ImportListExclusions/EditImportListExclusionModalContent.tsx
+++ b/frontend/src/Settings/ImportLists/ImportListExclusions/EditImportListExclusionModalContent.tsx
@@ -29,7 +29,6 @@ import styles from './EditImportListExclusionModalContent.css';
 
 const newImportListExclusion = {
   movieTitle: '',
-  movieYear: 0,
   type: 'scene',
   foreignId: '',
 };
@@ -80,7 +79,7 @@ function EditImportListExclusionModalContent({
   const { isFetching, isSaving, item, error, saveError, ...otherProps } =
     useSelector(createImportListExclusionSelector(id));
 
-  const { movieTitle, movieYear, foreignId, type } = item;
+  const { movieTitle, foreignId, type } = item;
 
   const dispatch = useDispatch();
   const previousIsSaving = usePrevious(isSaving);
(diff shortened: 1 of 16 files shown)


plz12345 (changes requested):
I had a feeling I misunderstood you.  Can you remove the brackets and foreign ID from Title?  Also, when you removed the Year column in the table, it broke alignment/spacing.  It is likely a mismatch in the column count vs. the header column count.
image

Regarding behavior, I tested by trying to add a scene via Scenes > Add Scene.

When it has a tag exclusion, the first click of the add button fails silently (400 Bad Request from the API), but the second seems to let it go through.

Log has a line
` [Warn] WhisparrErrorPipeline: Invalid request Tag(s): [Exclusion: [Tag][6458e5cf-4f65-400b-9067-582141e2a329][Virtual Reality]] excluded `

That's probably out of scope for this, as the front end isn't wired to receive errors, and the AddMovie method in the API conrtoller isn't set up to return them.

sampulsar:
The First time will add the exclusion to the scene, but if you are adding a scene manually you can add one that has been excluded. All of the automated lists, performer and studio refresh task will continue to exclude it. 

The layout has been corrected by fixing the table by correcting the colspan on the deleted as there are only 4 columns now.

I did format the movies, shown in my screenshot, but I did not push the commit.

plz12345:
OK, it didn't break anything new, either way, so as long as the table if fixed, I'm good.  We can circle back to fix UX on the behavior when we block an addition, later.  That's going to require local API changes and UI changes, since the API just returns an empty array on failure, not an error.

sampulsar:
Yeah, the majority of the time another one of the non-user add events would create the exclusion of the scene because it has been linked. 

The user would then search and the UI does display that the movie/scene has an exclusion and they then add the item, the tag/performer/studio checks are then skipped.

Adding one when there is not an exclusion has not been clean via the UI.

plz12345 (approved):
Looks good.  Thank you for wiring this up, end to end.

PR #14, opened 2026-01-18
New: Record Import Exclusion reasons

sampulsar:
#### Database Migration
YES - 0014 

#### Description
Record why an exclusion was added, via a delete, manunual, or automatically via another exclusion, or rule after date on studio. Can help determine why there is an exclusion on a scene.

Testing delete and exclude some items and then refresh studios and performer to see scenes exclusions being automatically created with the reason recorded. Can explain if someone deletes and excludes a performer because they do not want to see them in whisparr but does not understand why scenes are missing / excluded.

image

image

#### Todos
- [x] Tests

* Fixes whisparr/whisparr#946

### DIFF
--- a/src/NzbDrone.Core/ImportLists/ImportExclusions/ImportExclusionReason.cs
+++ b/src/NzbDrone.Core/ImportLists/ImportExclusions/ImportExclusionReason.cs
@@ -1,0 +1,12 @@
+namespace NzbDrone.Core.ImportLists.ImportExclusions
+{
+    public enum ImportExclusionReason
+    {
+        Manual,
+        Deleted,
+        PerformerExclusion,
+        StudioExcluded,
+        StudioAfterDate,
+        TagExclusion
+    }
+}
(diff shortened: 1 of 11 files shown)


src/NzbDrone.Core/ImportLists/ImportExclusions/ImportExclusionReason.cs
```diff
@@ -0,0 +1,12 @@
+namespace NzbDrone.Core.ImportLists.ImportExclusions
+{
+    public enum ImportExclusionReason
+    {
+        Manual,
+        Deleted,
```
plz12345: Recommend `DuringDelete` to avoid ambiguity.

src/NzbDrone.Core/ImportLists/ImportExclusions/ImportExclusionReason.cs
```diff
@@ -0,0 +1,12 @@
...
+        PerformerExclusion,
+        StudioExcluded,
```
plz12345: `StudioExclusion` to be consistent with the others

plz12345 (changes requested):

plz12345 (approved):

PR #20, opened 2026-01-18
Fix: Browser 401 redirect to login form

plz12345:
#### Database Migration
NO

#### Description
As reported on Discord, recent app updates broke the 401 redirect to the login page.  I found some quirks related to .NET 10 and the use of multiple middleware handlers (which we have), so ended up putting a shim in place to override the 401 if a higher layer set a Location (intended a redirect).  The API doesn't do this, so this should be safe for non-API requests (e.g., UI-only).

#### Todos
- [x] Tests

* Fixes whisparr/whisparr#947


sampulsar (approved):

plz12345:
Don't merge, I'm working on a patch to align the integration tests.

plz12345:
Will redo this, getting annoying with the tests.

PR #28, opened 2026-01-22
Fix: Performer API SQLite error on huge libraries

plz12345:
Added batching to the API calls for "all performers" and "all studios" to prevent SQLite from choking on parameter count.

#### Database Migration
NO

#### Todos
- [x] Tests

* Fixes #XXXX


sampulsar (approved):

sampulsar:
@plz12345 

This change also stops the one datalevel process for blocking other related reads for a shorter duration, eg allow an add or other related read to return without having to wait for all the performers to return.

It is hard to see the internals of SQLite to see what may be blocking at the row or table level.

PR #43, opened 2026-01-25
Fix: Update the backend Credit Controller

sampulsar:
#### Database Migration
YES - 0016

#### Description
Migrate more data into the credit table, so that the credit controller can be used.

Align the Credit Controller with the frontend.

Store a list of Performer Ids on the Metadata instead of the complete data. 

Some duplication of data Name, images on StashDB performers or the logic can be changed to save data.

The Cast Controller can return the Monitor fields if required.

Some Data can be removed once other changes happen, the new data is populated and other part change. There are comments if something is temporary.

The monitoring icons will only show if the performer can monitor. The state does not update correctly, so may be they will just be view only.

image

#### Todos
- [x] Tests

* Fixes whisparr/whisparr#318


sampulsar:
@plz12345 

Just realised that we should store the list of performer names on the movie so that the data is flattened on the movie.

sampulsar:
@plz12345 

Done. One the data is now denormalized it may make the paging and filtering easier.

src/NzbDrone.Core/Movies/Credits/Credit.cs
plz12345, line 1: @sampulsar Instead of storing performer images on both performer and credit, could we just use PerformerService to fetch the images per-credit, as needed?  It seems odd to me and will likely lead to image mismatches when a scene is refreshed but the performer isn't.

src/Whisparr.Api.V3/Credits/CreditResource.cs
```diff
@@ -14,39 +16,61 @@ public CreditResource()
...
             {
                 Id = model.Id,
+                PerformerId = model.Performer.Id,
                 MovieMetadataId = model.MovieMetadataId,
-                PerformerForeignId = model.PerformerForeignId,
-                PersonName = model.Performer?.Name,
+                ForeignId = model.PerformerForeignId,
+                PersonName = model.PersonName,
                 Character = model.Character,
+                Images = model.Images,
```
plz12345, line 112: @sampulsar Since this is the API's layer for this, could we just fetch Performer.Images on-demand via `PerformerService.FindByForeignId`, here, instead of storing images on the Credit?  Same for Names, really, it just seems weird to store them in two places, both of which can be refreshed separately.

src/Whisparr.Api.V3/Movies/MovieResource.cs
```diff
@@ -79,6 +79,7 @@ public MovieResource()
         public Ratings Ratings { get; set; }
         public MovieFileResource MovieFile { get; set; }
         public MovieCollectionResource Collection { get; set; }
+        public List<string> PerformerForeignIds { get; set; }
```
plz12345, line 82: @sampulsar is this for backward compatibility?  The Credits would already have those same foreign ID's.

src/NzbDrone.Core/Movies/Credits/Credit.cs
sampulsar: Not all credits are linked to a performer. The Scene metadata from skyhook has the performer. The images are never a mismatch as there is mapping to the local file on the system. It will always be the same image for these performers, as it is just a reference to an image, and not the data.

Only credits outside of whisparr, TMDB, TDBP performers will get the image directly from there sites.

FYI: the coverMapper does this within the controllers to stop any image mismatch. It will get the image locally if it has been downloaded, otherwise the true source URL is used.

_coverMapper.ConvertToLocalUrls(0, resource.Images);

src/Whisparr.Api.V3/Credits/CreditResource.cs
sampulsar: Not all Credits are for performers within Whisparr. The credit is a light weight version of a performer that has enough data to do just what it needs.

src/Whisparr.Api.V3/Movies/MovieResource.cs
sampulsar: This is to help migrate away from the credits object on this resource as the performer page is doing some filtering of all the movies to find ones for a performer.

This is flattened data to help with this, it can only be used once populated.

It is easier to let the change go out for a couple of days and then alter the fronted and remove the large credit object from this resource.

It is a way out slowly altering Whisparr, without having to do complex migration of data. It also breaks the change into smaller manageable changes.

src/NzbDrone.Core/Movies/Credits/Credit.cs
sampulsar: You may only save limited data as this is the image in that database.

[
  {
    "coverType": "headshot",
    "remoteUrl": "https://stashdb.org/images/7fe08a9e-5a4e-4ae3-9f9e-820a86f27304"
  }
]

sampulsar:
@plz12345 

Change the data service, just in case an image or name is changed on the performer, to handle any time difference between the refresh of the scene and performer. The performer is the system record, if it is available, as there are non-stashdb credits.

// Use Performer Data over a Credit version if available
if (performer?.Images.Count > 0)
{
    credit.Images = performer.Images;
}

sampulsar:
@plz12345 

This can be reviewed again. I can not see any additional comments

plz12345 (approved):

PR #54, opened 2026-01-28
Fix: StashDB Import List fields not showing existing values

plz12345:
This is a byproduct of upstream refactoring of the SelectOptions definition (I believe).  The UI is expecting to work with enum int values, but this import list type is working with strings.  The strings were also stored in the DB that way.  This doesn't align with other import lists' settings.  In order to avoid a DB migration, I opted to fix the casting in the places these are used.

Going forward, the settings will be saved in line with *arr standards, but existing lists values already saved will be translated to avoid runtime errors.  If a user re-saves an existing list, it will be corrected to store the appropriate settings shape.

### To Test
- Back up DB
- Change docker tag to `ghcr.io/hotio/whisparr:v3-pr-fix-stashdb-import-list`
- Start container

Expected result
- UI should load normally
- Import lists should load when edited, fields like sort descening should show the existing values
- New lists should be add-able
- Re-saving an existing list should update the Settings column to store select list enums as int instead of string
- Saving a list should successfully trigger a refresh

#### Database Migration
NO

#### Todos
- [x] Tests

* Fixes whisparr/whisparr#948

### DIFF
--- a/src/NzbDrone.Core/Datastore/Migration/017_migrate_stashdb_enum_fields.cs
+++ b/src/NzbDrone.Core/Datastore/Migration/017_migrate_stashdb_enum_fields.cs
@@ -1,0 +1,90 @@
+using System;
+using System.Text.Json;
+using FluentMigrator;
+using NzbDrone.Core.Datastore.Migration.Framework;
+using NzbDrone.Core.ImportLists.StashDB;
+
+namespace NzbDrone.Core.Datastore.Migration
+{
+    [Migration(017)]
+    public class migrate_stashdb_enum_fields : NzbDroneMigrationBase
+    {
+        protected override void MainDbUpgrade()
+        {
+            // Read all relevant rows into memory first, then update as needed (avoids holding SELECT open during UPDATE)
+            Execute.WithConnection((conn, tran) =>
+            {
+                var importLists = new System.Collections.Generic.List<(int Id, System.Text.Json.Nodes.JsonObject SettingsObj)>();
+                using (var cmd = conn.CreateCommand())
+                {
+                    cmd.CommandText = "SELECT \"Id\", \"Settings\", \"Implementation\" FROM \"ImportLists\" WHERE \"Implementation\" LIKE 'StashDB%'";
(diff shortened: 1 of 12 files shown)


plz12345:
@sampulsar , latest build should fix that.  I also fixed the issue (again) that was causing date fields to not show the calendar picker.  It's an upstream bug, and I missed it when doing upstream merges.

NVM, it doesn't pass testing on existing lists, now.  This is such a pain to avoid DB migration...

src/NzbDrone.Core/Datastore/Migration/017_migrate_stashdb_enum_fields.cs
```diff
@@ -0,0 +1,90 @@
...
+                    changed |= ConvertEnumField(settingsObj, "studiosFilter", typeof(FilterModifier), id);
+                    changed |= ConvertEnumField(settingsObj, "tagsFilter", typeof(FilterModifier), id);
+                    changed |= ConvertEnumField(settingsObj, "filter", typeof(FavoriteFilter), id);
+
+                    if (changed)
+                    {
+                        var newJson = settingsObj.ToJsonString(new JsonSerializerOptions { WriteIndented = true });
+                        using (var updateCmd = conn.CreateCommand())
+                        {
+                            updateCmd.CommandText = "UPDATE \"ImportLists\" SET Settings = @settings WHERE Id = @id";
```
sampulsar: Missing the quotes around the field names

I alter to what it should be, but will wait for the PR to update to run the final conversion. 

Image

plz12345: Thanks, I need to stop doing high-precision coding stuff when my brain isn't in that mode.  Updating now.

sampulsar (approved):

sampulsar:
@plz12345 

Tested in Postgres and it migrated the data.

plz12345:
Awesome, thank you for that.  I may set up a test env for PGSQL at some point, but glad you were able to confirm.

PR #56, opened 2026-01-28
Chore: Remove Credits data form MovieResource

sampulsar:
#### Database Migration
YES - 018

#### Description
This change will make the client do less work as the data is processed in a simply flat way for the cast. For the movieResource the name and the StashId was the two main parts that where used.

I went with a denormalized structure so that those common items are stored on with the movieMetaData instead of reading via the credit service everytime. The stashIds, performerForiegnIds, field can be removed at a later date if paging is implemented. 

This implementation would call ListByPerformerForeignId on the movieController to get the ids for the performer and then pass that to bulk on the movie controller to get the details of the performers movies.

Remove the large Credit object from the Metadata object.
Replaced with two lists, performer names and the performers StashIds.

3 Commits
-Backend
-UnitTest
-Frontend

Note: A movie refresh is required as the two lists where not populated in the last commit.

Changes the Studio and Performer Details pages
Changed the Add Movie page, a gender icon is no longer displayed as we are passing just the names

Add Scene/Movie
image
Original with Gender (Only on this page for a scene/movie)
image

Performer
image

Studio 
image

#### Todos
- [x] Tests

* Fixes #XXXX


frontend/src/AddMovie/AddNewMovie/AddNewMovie/AddNewMovieSearchResult.js
```diff
@@ -23,25 +23,6 @@ import translate from 'Utilities/String/translate';
 import AddNewMovieModal from './AddNewMovieModal';
 import styles from './AddNewMovieSearchResult.css';
 
-function genderToIcon(gender) {
```
plz12345, line 26: Why remove this?  I spent a decent amount of time making these performer labels distinct with the gender icons... :/

src/NzbDrone.Core/Movies/AddMovieService.cs
```diff
@@ -309,7 +309,7 @@ private Movie SetPropertiesAndValidate(Movie newMovie)
                     }
                 }
 
-                var performerForeignIds = newMovie.MovieMetadata.Value.Credits.Select(c => c.PerformerForeignId);
+                var performerForeignIds =  newMovie.MovieMetadata.Value.PerformerForeignIds;
```
plz12345: the extra space before `newMovie` is going to trigger a diff the next time one of us edits this and the IDE fixes it.

sampulsar:
@plz12345 

The Search results will return Cast Info via the Credit, but it is now named differently so it does not get used, by tests, and other sections. 

image

frontend/src/AddMovie/AddNewMovie/AddNewMovie/AddNewMovieSearchResult.js
sampulsar: Restored

plz12345:
I fired it up locally, and it works well.  I did run into the issue where you need to refresh everything.  For my testing, I purged the dev DB of all performers, studios, and scenes, and added a few.  No issues came up.

sampulsar:
I may not have run my Dev environment long enough with the last change to Credits to populate data, but the refresh worked and populated that data.

plz12345 (approved):

PR #65, opened 2026-02-02
Fix: Correct performer naming for existing items

sampulsar:
#### Database Migration
NO

#### Description
Currently a difference between the Skyhook data and a credit existing item.

Grab the credit data and fallback to skyhook if it is a new item.

#### Todos
- [x] Tests

* Fixes whisparr/whisparr#991

### DIFF
--- a/src/NzbDrone.Core/MetadataSource/SkyHook/Resource/CreditsResource.cs
+++ b/src/NzbDrone.Core/MetadataSource/SkyHook/Resource/CreditsResource.cs
@@ -10,6 +10,7 @@ namespace NzbDrone.Core.MetadataSource.SkyHook.Resource
     public class CastResource
     {
         public int Order { get; set; }
+        public string PersonName { get; set; }
         public string Character { get; set; }
         public string CreditId { get; set; }
         public PerformerResource Performer { get; set; }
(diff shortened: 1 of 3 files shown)


sampulsar:
Going to rework this so that the data from Skyhook is correct, and while that happens I am going to map the Performer.Name to PersionName.

The data from skyhook should be in the same structure as Whisparr. Make it simple.

src/NzbDrone.Core/MetadataSource/SkyHook/Resource/CreditsResource.cs
```diff
@@ -10,6 +10,7 @@ public class Credits
     public class CastResource
     {
         public int Order { get; set; }
+        public string PersonName { get; set; }
```
sampulsar: Add New Field that should be added to SkyHook, to make both Skyhook and Credit the same structure

src/NzbDrone.Core/MetadataSource/SkyHook/SkyHookProxy.cs
```diff
@@ -1314,7 +1314,7 @@ private static Credit MapCast(CastResource arg)
         {
             var newActor = new Credit
             {
-                PersonName = arg.Performer.Name,
+                PersonName = arg.PersonName.IsNotNullOrWhiteSpace() ? arg.PersonName : arg.Performer.Name,
```
sampulsar: Use the Person name if populated, likely not and then fallback to the old performer name

plz12345 (approved):

PR #70, opened 2026-02-03
Fix: Harden Metadata API response handling

plz12345:
#### Database Migration
NO

#### Description
SkyHook’s API response processing mapping methods assumed perfect responses, suitable for single-item queries but inadequate for lists.  If any item fails to map, the whole thing fails. To address this, SkyHookProxy now employs a dual approach:

- Added explicit null-checks for mapping methods that use classes as input parameters.  This solves any potential NRE's when binding properties from the input param.
- Moved away from LINQ `SelectList` iterators in favor of per-item `try...catch `when iterating a list.  This combined with the null-check throws should result in both more predictable results and better exception logging.

Also added a suggestion in .editorconfig to surface unused method params.  It them appear "dim" and they show a tooltip on hover.  See screenshot on PR.

#### Screenshot
image

#### Todos
- [x] Tests

- Fixes whisparr/whisparr#994
- Fixes whisparr/whisparr#995

### DIFF
--- a/src/NzbDrone.Core/MetadataSource/SkyHook/SkyHookProxy.cs
+++ b/src/NzbDrone.Core/MetadataSource/SkyHook/SkyHookProxy.cs
@@ -506,7 +506,7 @@ namespace NzbDrone.Core.MetadataSource.SkyHook
             var tpdbIds = httpResponse.Resource.TpdbMovies;
             var tmpdIds = new List<int>();
 
-         // Check while Skyhook is transitioning to return TMDB company movies.
+            // Check while Skyhook is transitioning to return TMDB company movies.
             if (httpResponse.Resource.Movies != null)
             {
                 tmpdIds = httpResponse.Resource.Movies.ConvertAll(int.Parse);
@@ -529,6 +529,11 @@ namespace NzbDrone.Core.MetadataSource.SkyHook
 
         public MovieMetadata MapMovie(MovieResource resource)
         {
+            if (resource == null)
+            {
+                throw new ArgumentNullException(nameof(resource));
+            }
+
             var movie = new MovieMetadata();
             var altTitles = new List<AlternativeTitle>();
             var metadataSource = MapMetadataSource(resource.ItemType, resource.ForeignIds);
@@ -632,6 +637,11 @@ namespace NzbDrone.Core.MetadataSource.SkyHook
 
         public MovieMetadata MapMovieToTmdbMovie(MovieMetadata movie)
         {
+            if (movie == null)
+            {
+                throw new ArgumentNullException(nameof(movie));
+            }
+
             try
             {
                 var newMovie = movie;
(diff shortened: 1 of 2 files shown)


sampulsar (approved):

sampulsar:
@plz12345 

There may need to be additional lines to ensure that the PersonName is populated when the data from SkyHook is 

RefreshMovieService.cs

private Movie RefreshMovieInfo(int movieId)

...

            // Update the flattened performer lists on the movie metadata
            performerList.ForEach(performer => movieMetadata.PerformerForeignIds.Add(performer.ForeignId));

            **// Make sure credit person names are not null
            creditInfo.Map(c => c.PersonName = string.IsNullOrWhiteSpace(c.PersonName) ? c.Performer.Name : c.PersonName);
            creditInfo.Map(c => c.PersonName = string.IsNullOrWhiteSpace(c.PersonName) ? string.Empty: c.PersonName);**

            creditInfo.ForEach(credit => movieMetadata.PerformerNames.Add(credit.PersonName));

src/NzbDrone.Core/MetadataSource/SkyHook/SkyHookProxy.cs
```diff
@@ -1438,62 +1512,79 @@ private static Gender MapGender(string gender)
...
+            {
+                throw new ArgumentNullException(nameof(castResource));
+            }
+
             var newActor = new Credit
             {
-                PersonName = arg.Performer.Name,
-                Character = arg.Character,
-                Order = arg.Order,
+                PersonName = castResource.Performer.Name,
```
sampulsar, line 1574: Should do this the same as MapCredit, as the metadata will eventually be the same structure for both.


PersonName = castResource.PersonName.IsNotNullOrWhiteSpace() ? castResource.PersonName : castResource.Performer.Name,

sampulsar (changes requested):

src/NzbDrone.Core/MetadataSource/SkyHook/SkyHookProxy.cs
sampulsar: May also need to do both there to also handle if both are null. 

Or 

within make sure that the value is not null, as the database is not nullable. There bad credits would need to be removed,by a filter, ot set to an empty string.

RefreshMovieService.cs

private Movie RefreshMovieInfo(int movieId)

creditInfo.Map(c => c.PersonName = string.IsNullOrWhiteSpace(c.PersonName) ? string.Empty: c.PersonName);

plz12345:
@sampulsar I wasn't proposing refactoring anything outside of the SkyHookProxy class as it is currently generating NRE's and those can be prevented.  My goals on this targeted edit were:

1. Prevent NRE's in the mapping methods in SkyHookProxy
2.Gracefully handle inconsistent v4 API data in list-type responses.

sampulsar (approved):

PR #76, opened 2026-02-05
New: Add additional performer fields (light)

sampulsar:
#### Database Migration
YES - 0019

#### Description
Lighter weight change to the FE for Additional performer data.

Allows for additional improvements to happen before the API exposes all of the new data in the database.

image

image

#### Todos
- [x] Tests

* Fixes #XXXX


src/NzbDrone.Core/MetadataSource/SkyHook/SkyHookProxy.cs
```diff
@@ -1381,9 +1381,16 @@ private static Credit MapCast(CastResource castResource)
                 throw new ArgumentNullException(nameof(castResource));
             }
 
+            var personName = castResource.PersonName.IsNotNullOrWhiteSpace() ? castResource.PersonName : castResource.Performer.Name;
```
plz12345:
```
var personName = castResource.PersonName.IsNotNullOrWhiteSpace()
    ? castResource.PersonName
    : castResource.Performer?.Name ?? string.Empty;
```

If performer is non-null, the single ? isn't needed.

src/NzbDrone.Core/Movies/Performers/Performer.cs
```diff
@@ -73,6 +87,24 @@ public enum Ethnicity
         Other
     }
 
+    public enum BreastTypeEnum
+    {
+        Natural,
+        Fake,
```
plz12345, line 93: Can we use `Augmented` instead?

src/Whisparr.Api.V3/Performers/PerformerResource.cs
```diff
@@ -126,6 +136,13 @@ public static PerformerResource ToResource(this Performer model)
...
             }
 
+            // Extract the name with the Disambiguation in the same format as StashDB (For User Display)
+            var fullname = model.Name;
+            if (model.Disambiguation.IsNotNullOrWhiteSpace())
+            {
+                fullname = $"{fullname} ({model.Disambiguation})";
+            }
+
             return new PerformerResource
```
plz12345, line 146: You seem to be missing a lot of the new fields you added to the database, here in the API layer.

plz12345 (changes requested):

src/Whisparr.Api.V3/Performers/PerformerResource.cs
sampulsar: Yes, this is the light version, that alters the FE minimally, so the data is populated and can be changed in the future, next month when more performers have data.

The UI can be improved and then those properties can be added to the resource.

src/NzbDrone.Core/Movies/Performers/Performer.cs
sampulsar: These are the Types from StashDB, but the client can translate Fake to "Augmented".

plz12345: I didn't realize that.  So they are translating that, too.

PR #84, opened 2026-02-07
Fix: Read the Gender value for Credit Performer to correct naming

sampulsar:
#### Database Migration
YES - 021

#### Description
Read all the credit data, for naming tokens. 
Handle old data for the credit personname

Second Commit Checks information for Scene Impot
Handle when there are no root folders for a scene Import
Checks that the Identification has a Movie with a movie ID.

Third
Check the profileID, so it does not throw and exception, the first will then be used.

Fourth
Set a Fallback Quality Profile, so that it can correct issues with Quality Profiles.

image

image

#### Todos
- [x] Tests

* Fixes #XXXX
* refs https://github.com/Whisparr/Whisparr-Eros/issues/781

### DIFF
--- a/src/NzbDrone.Core/Movies/MovieRepository.cs
+++ b/src/NzbDrone.Core/Movies/MovieRepository.cs
@@ -156,7 +156,12 @@ namespace NzbDrone.Core.Movies
                 {
                     movie.MovieMetadata = metadata;
                     movie.MovieFile = file;
-                    movie.QualityProfile = qualityProfiles[movie.QualityProfileId];
+
+                    // Check Quality Key exists
+                    if (qualityProfiles.ContainsKey(movie.QualityProfileId))
+                    {
+                        movie.QualityProfile = qualityProfiles[movie.QualityProfileId];
+                    }
 
                     if (alternativeTitles.TryGetValue(movie.MovieMetadataId, out var altTitles))
                     {
--- a/src/NzbDrone.Core/Movies/MovieRepository.cs
+++ b/src/NzbDrone.Core/Movies/MovieRepository.cs
@@ -158,10 +158,14 @@ namespace NzbDrone.Core.Movies
                     movie.MovieFile = file;
 
                     // Check Quality Key exists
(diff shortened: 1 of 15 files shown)


src/NzbDrone.Core/Movies/MovieRepository.cs
```diff
@@ -156,7 +156,12 @@ public override IEnumerable<Movie> All()
                 {
                     movie.MovieMetadata = metadata;
                     movie.MovieFile = file;
-                    movie.QualityProfile = qualityProfiles[movie.QualityProfileId];
+
+                    // Check Quality Key exists
+                    if (qualityProfiles.ContainsKey(movie.QualityProfileId))
```
plz12345: @sampulsar can we set a default here, if the condition would go to an else branch?

plz12345 (changes requested):
Just one change to tweak setting a quality profile of last resort

src/NzbDrone.Core/Movies/MovieRepository.cs
sampulsar: What is the default? 
That should be in the client, or is it the first quality Profile?
What it the client default is deleted?

If the ID is not within the database the the item would not have a quality, eg null. It would need to be corrected manually as the intended profile can not be determined if profiles have been deleted.

plz12345: The QualityProfileId is non-nullable in the database, so we should protect against that anywhere we discover or handle for nulls on it.  Better to create a movie with a default than throw a constraint error into the logs.  As for defaults, I would suggest putting a fluent rule to prevent deletion of the default one, if not there already, then using that index/enum as the default for this.

sampulsar: The value is not null, just a number that is not in QualityProfiles, that is the containsKey check.

The quality profile can be deletes if it is not used, but the id can be passed and uses as it is just an Id. The Id may also be 0, as that is the default initialization.

plz12345 (approved):

PR #108, opened 2026-02-13
New: Import process over the import directory

sampulsar:
#### Database Migration
YES - 022

Description
Readded as other branch was corrupted.

Use most of the radarr import process.
Correct the search to parse the filenames and then search.
Display the search results formatted for scenes and movies.
import and move the filenames correctly.

#### Screenshot (if UI related)
image
image
image

#### Todos
- [x] Tests
- [x] Translation Keys (./src/NzbDrone.Core/Localization/Core/en.json)
- [x] [Wiki Updates](https://wiki.servarr.com)

#### Issues Fixed or Closed by this PR

* Fixes #XXXX

PR #118, opened 2026-02-14
Fix: Give V3 a unique Key for browser data so it works alongside V2

sampulsar:
#### Database Migration
NO

#### Description
V2 and V3 used the same persistent key that is used to store data in local storage.

This does cause issues when some one changes form v2 to v3 as there is already values, usually when they add some items as a dropdown will have a value that does not exist in V3.

Best to separate the data with a different key. 

Downside for all users is any existing settings are lost, and the defaults are applied. 
Upside is that the correct defaults for all items are used and they do not need to be manually cleared.

This gives a known starting point for the data. 

Can this go in after the import change #108 

#### Todos
- [x] Tests

* Fixes #XXXX

### DIFF
--- a/frontend/src/Store/Middleware/createPersistState.js
+++ b/frontend/src/Store/Middleware/createPersistState.js
@@ -96,7 +96,7 @@ function merge(initialState, persistedState) {
   return computedState;
 }
 
-const KEY = 'WhisparrV3';
+const KEY = 'WhisparrEros';
 
 const config = {
   slicer,
--- a/frontend/src/Store/Middleware/createPersistState.js
+++ b/frontend/src/Store/Middleware/createPersistState.js
@@ -96,7 +96,7 @@ function merge(initialState, persistedState) {
   return computedState;
 }
 
-const KEY = 'whisparr';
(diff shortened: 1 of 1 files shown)


frontend/src/Store/Middleware/createPersistState.js
```diff
@@ -96,7 +96,7 @@ function merge(initialState, persistedState) {
   return computedState;
 }
 
-const KEY = 'whisparr';
+const KEY = 'WhisparrV3';
```
plz12345: @sampulsar Can you call this `WhisparrEros` instead?  V3 permanently locks us into that major version.

How does this impact existing v3 users?

plz12345 (changes requested):

frontend/src/Store/Middleware/createPersistState.js
sampulsar: If a user have a Quality profile set or any exclude on delete setting, this change will reset those values. 

It is the same as if they started using it on a new device/browser.

Change to WhisparrEros

plz12345 (approved):

PR #123, opened 2026-02-16
New: Add Performer Country Column and Filter

sampulsar:
#### Database Migration
NO

#### Description
Add Performer Country as a field and filter 

Add translation and label for the flags

The Code is listed as that is the sort order.

image
image

#### Todos
- [x] Tests

* Fixes #XXXX


frontend/src/Performer/Index/Table/PerformerIndexRow.tsx
```diff
@@ -145,6 +148,16 @@ function PerformerIndexRow(props: PerformerIndexRowProps) {
       );
       return;
     }
+    if (name === 'country') {
+      cells.push(
+        <TableRowCell key={name} className={styles[name]}>
+          {countryCodeName === country
+            ? country
+            : `${country} (${countryCodeName})`}
```
plz12345: @sampulsar can you change this so it shows the code only, and puts the full country name in the title/tool tip?  The width is a bit excessive, IMO.  Also, does `countryCodeName` get localized automatically?  I don't see any localization here, so assuming either no, or it's buried in that helper.

plz12345 (changes requested):

frontend/src/Performer/Index/Table/PerformerIndexRow.tsx
sampulsar: the translation is within i18n-iso-countries, as that has the list of all the countries in with the translations.

image

plz12345 (approved):

PR #124, opened 2026-02-16
New: Add Library Statistics to Status page

sampulsar:
#### Database Migration
NO

#### Description
Add statistics about the library.

Simple Counts of scenes, movies, studios and performers

image

#### Todos
- [ ] Tests
- [ ] [Wiki Updates](https://wiki.servarr.com)

* Fixes #XXXX


src/Whisparr.Api.V3/System/SystemController.cs
```diff
@@ -88,7 +100,11 @@ public SystemResource GetStatus()
                 PackageVersion = _deploymentInfoProvider.PackageVersion,
                 PackageAuthor = _deploymentInfoProvider.PackageAuthor,
                 PackageUpdateMechanism = _deploymentInfoProvider.PackageUpdateMechanism,
-                PackageUpdateMechanismMessage = _deploymentInfoProvider.PackageUpdateMechanismMessage
+                PackageUpdateMechanismMessage = _deploymentInfoProvider.PackageUpdateMechanismMessage,
+                MovieCount = _movieMetadataService.CountByType(ItemType.Movie),
```
plz12345, line 104: @sampulsar this will have performance hits on large libraries.

Each repo class inherits from `BasicRepository`, which implements an efficient `Count()` method.  You can use that and call it through the other layers, so it's not fetching a ton of unused object data for each row of this.

As a prop
```csharp
public int Count => _performerRepository.Count();
```

or as a method
```csharp
public int Count()
{
    return _performerRepository.Count();
}
```
You'd just incorporate that in PerformerService so you can grab it from this or any other controller.

The movie vs. scene bit may require a new MovieRepository method that adds the where clause on itemType, but will still be more resource-efficient than what you have.

plz12345 (changes requested):

src/Whisparr.Api.V3/System/SystemController.cs
sampulsar: Change the query to use a statement that will count using a database index.

plz12345 (approved):

PR #129, opened 2026-02-20
New: Improve Import item type and matching accuracy

sampulsar:
#### Database Migration
NO

#### Description
Create paths form import movie and scene for each folder
Set the scan for types based on the path
Clean the extension form a filename for a search.

image
image

#### Todos
- [x] Tests

- Fixes whisparr/whisparr#1015
- Fixes whisparr/whisparr#1014
- Fixes whisparr/whisparr#1005

### DIFF
--- a/frontend/src/AddMovie/ImportMovie/Import/SelectMovie/ImportMovieSelectMovieConnector.js
+++ b/frontend/src/AddMovie/ImportMovie/Import/SelectMovie/ImportMovieSelectMovieConnector.js
@@ -34,7 +34,15 @@ class ImportMovieSelectMovieConnector extends Component {
     const isMovies = window.location.pathname.includes('movies');
     const isScenes = window.location.pathname.includes('scenes');
     // eslint-disable-next-line no-nested-ternary
-    const itemType = isMovies ? 'movie' : isScenes ? 'scene' : null;
+    const itemType = (() => {
+      if (isMovies) {
+        return 'movie';
+      }
+      if (isScenes) {
+        return 'scene';
+      }
+      return null;
+    })();
 
     this.props.queueLookupMovie({
       name: this.props.id,
--- a/frontend/src/AddMovie/ImportMovie/Import/SelectMovie/ImportMovieSelectMovieConnector.js
+++ b/frontend/src/AddMovie/ImportMovie/Import/SelectMovie/ImportMovieSelectMovieConnector.js
@@ -31,9 +31,15 @@ class ImportMovieSelectMovieConnector extends Component {
   // Listeners
 
   onSearchInputChange = (term) => {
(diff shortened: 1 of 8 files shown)


frontend/src/AddMovie/ImportMovie/Import/SelectMovie/ImportMovieSelectMovieConnector.js
```diff
@@ -31,9 +31,15 @@ class ImportMovieSelectMovieConnector extends Component {
   // Listeners
 
   onSearchInputChange = (term) => {
+    const isMovies = window.location.pathname.includes('movies');
+    const isScenes = window.location.pathname.includes('scenes');
+    // eslint-disable-next-line no-nested-ternary
```
plz12345, line 36: No need for the linter escaping.  Let's be honest, scene is going to be more common, so defaulting to scene is more practical than defaulting to movie.  I presume passing `null` would default to movie in the controller, so we can prevent that here rather than changing code in parser.

```typescript
const itemType = isMovies ? 'movie' : 'scene';
```

plz12345 (changes requested):

frontend/src/AddMovie/ImportMovie/Import/SelectMovie/ImportMovieSelectMovieConnector.js
sampulsar: There is only one import folder for both scenes and movies.

There is three navigation paths, 
Scenes >> Import Scene (Search Context is scenes, the user has some intent to import scenes)
Movies >> Import Movie (Search Context is movie, the user has some intent to import scenes)
 Settings >> Media Management (Search context is not known, so the file name needs to determine if it is a scene of movie)

sampulsar: When null is passed the filename is parsed, so a standard scene file name will {studio} {date} {???} will be a scene,

A pure title is usually considered a movie, as that is how the index search parse the information. Take Note Charlie Red looks like a movie title.

The parsing can also clean some parts from the filename, that something like the stashdb search does not like, eg release group, quality etc.

plz12345: Right, but if it isn't a "standard scene name", it defaults to movie, regardless of the intent the user had.  That was my point.  So as long as the URL part works as intended, I guess we're OK with not forcing a non-null default.

Maybe change to:

```typescript
const itemType = (() => {
  if (isMovies) return 'movie';
  if (isScenes) return 'scene';
  return null;
})();
```

So you don't need the lint escape comment.

plz12345 (approved):

PR #131, opened 2026-02-20
New: Migrate MovieDetails to React Query

plz12345:
#### Database Migration
NO

#### Description
This is the first part of a two-part phase to remove Movie items from Redux, to boost web browser perormance in the app when managing large (5-figure) and huge (6-figure) libraries.

- Migrates the Movie Details page from a Redux-connected class component to a functional component powered by React Query, removing MovieDetailsConnector.js and MovieDetailsPageConnector.js entirely.
- Adds new useMovie(), useMovieIndexQuery(), useMoviesByForeignIds(), and useToggleMovieMonitored() hooks in useMovie.ts, replacing the previous Redux selector-based approach.
- New React hooks for data fetching and some editing. Does not replace Redux for every command (refresh, search, etc.).
- Updates SignalRConnector.js to invalidate React Query cache when videos are updated over the WebSocket, keeping the UI in sync without Redux.  Similar pattern to the performer and studio implementations.
- Removed MovieFile items from Redux store, this is purely React Query for fetching, now.
- Modularized MovieDetails page API calls by section, to allow for better async loading of sections on slower networks.  The deteails, files, titles, and cast all load async from each other.
- Fixed multiple bugs in image caching/loading, especially when using movie credits.  Images should load from local cache more often, only falling back to hotlinking if the image isn't actually stored (next refresh should get it).

#### Backend additions to support the frontend migration
- New `GET /api/v3/movie/{id}` endpoint that resolves both numeric IDs and foreign IDs (TPDB/Stash), replacing the need for pre-loaded Redux state on page load (Redux will be removed when Movie/Scene Index gets pagination)
- New `POST /api/v3/movie/list` endpoint for bulk lookup by a list of foreign IDs.
- New `POST /api/v3/movie/paged` endpoint for paginated, filtered, and sorted movie index results.
- Adds `FindByForeignIds` to `MovieRepository` and `MovieService`.

No major UI changes, all background stuff.

#### Todos
- [x] Tests

- Refs whisparr/whisparr#961


frontend/src/Movie/Details/MovieDetails.tsx
```diff
@@ -106,573 +72,481 @@ interface Props {
...
+    handleInteractiveImportPress,
+    handleInteractiveImportModalClose,
+    handleEditMoviePress,
+    handleEditMovieModalClose,
+    handleDeleteMoviePress,
+    handleDeleteMovieModalClose,
+  } = useMovieDetailsModals();
+
+  // If loading or error, show appropriate UI
+  if (isLoading) return <div>Loading...</div>;
```
sampulsar: Should show the loading indicator instead of just text

```
return (
        <PageContent>
          <LoadingIndicator />
        </PageContent>
      );
```

sampulsar (changes requested):

frontend/src/Movie/Details/MovieDetails.tsx
plz12345: Good catch.  Fixed that, and the indicators for error and loading on cast/credits as well as movie files.  Extra files is still in Redux, but will be migrated in the next PR, at which point I"ll cover that section in the same way.

sampulsar (approved):

PR #132, opened 2026-02-21
Fix: Show Movie Monitor Toggle

sampulsar:
#### Database Migration
NO

#### Description
Correct the checks to display the movie monitor toggle if it can be monitored.

Hide if it is none, or check that the source has a populated key value.

The case was incorrect for the source. 

#### Todos
- [x] Tests

* Fixes #XXXX

### DIFF
--- a/frontend/src/Performer/Index/usePerformerIndex.ts
+++ b/frontend/src/Performer/Index/usePerformerIndex.ts
@@ -18,8 +18,8 @@ import {
   setPerformerTableOption,
   setPerformerView,
 } from 'Store/Actions/performerActions';
+import { fetchGeneralSettings } from 'Store/Actions/Settings/general';
 import { createCustomFiltersSelector } from 'Store/Selectors/createClientSideCollectionSelector';
-import { useGeneralSettings } from 'Studio/Details/useStudioDetails';
 import { usePerformerIndexQuery } from './usePerformerIndexQuery';
 
 /**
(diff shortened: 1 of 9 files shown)


frontend/src/Performer/Index/usePerformerIndex.ts
```diff
@@ -19,6 +19,7 @@ import {
   setPerformerView,
 } from 'Store/Actions/performerActions';
 import { createCustomFiltersSelector } from 'Store/Selectors/createClientSideCollectionSelector';
+import { useGeneralSettings } from 'Studio/Details/useStudioDetails';
```
plz12345: @sampulsar can you just copy useGeneralSettings from `usePerformerDetails` to `usePerformerIndex`?  It's  a small function, and keeps the data domain/boundary in place so we're not mixing code across distinct objects.

frontend/src/Studio/Index/useStudioIndex.ts
```diff
@@ -4,6 +4,7 @@ import { useHistory } from 'react-router-dom';
 import ModelBase from 'App/ModelBase';
 import { useSelect } from 'App/SelectContext';
 import AppState from 'App/State/AppState';
+import { useGeneralSettings } from 'Performer/Details/usePerformerDetails';
```
plz12345: @sampulsar can you copy `useGeneralSettings` into `useStudioIndex` instead?  Not sure why on this one you used the one in `usePerformerDetails` and on the `usePerformerIndex` use used the one in `useStudioDetails`, but crossing data boundaries for this is problematic on dependencies.

plz12345 (changes requested):

plz12345 (approved):

PR #133, opened 2026-02-21
Fix: Navigate using the titleSlug

sampulsar:
#### Database Migration
NO

#### Description
Calculate a titleSlug for TMDB that is not numeric, so that the correct API endpoint is hit.

Use the titleSlug for navigation, so that the external Id is displayed for all items.

MovieTitleLink is now used again for all navigations to format and the link description and href.

another potential way to fix, would be to calculate tmdb title slug to match the TMDB site.

https://www.themoviedb.org/movie/699665-consumed-by-desire

They use a title slug that is the id and then the title.
699665-consumed-by-desire

#### Todos
- [ ] Tests
- [ ] [Wiki Updates](https://wiki.servarr.com)

* Fixes #XXXX

### DIFF
--- a/frontend/src/Movie/Index/Table/MovieIndexRow.tsx
+++ b/frontend/src/Movie/Index/Table/MovieIndexRow.tsx
@@ -4,7 +4,6 @@ import { useSelect } from 'App/SelectContext';
 import { MOVIE_SEARCH, REFRESH_MOVIE } from 'Commands/commandNames';
 import Icon from 'Components/Icon';
 import IconButton from 'Components/Link/IconButton';
-import Link from 'Components/Link/Link';
 import SpinnerIconButton from 'Components/Link/SpinnerIconButton';
 import MovieTagList from 'Components/MovieTagList';
 import RelativeDateCell from 'Components/Table/Cells/RelativeDateCell';
@@ -19,6 +18,7 @@ import MovieDetailsLinks from 'Movie/Details/MovieDetailsLinks';
 import EditMovieModal from 'Movie/Edit/EditMovieModal';
 import createMovieIndexItemSelector from 'Movie/Index/createMovieIndexItemSelector';
 import { Statistics } from 'Movie/Movie';
+import MovieTitleLink from 'Movie/MovieTitleLink';
 import { executeCommand } from 'Store/Actions/commandActions';
 import createUISettingsSelector from 'Store/Selectors/createUISettingsSelector';
 import { SelectStateInputProps } from 'typings/props';
@@ -50,6 +50,7 @@ function MovieIndexRow(props: MovieIndexRowProps) {
   const {
     monitored,
     foreignId,
+    titleSlug,
     title,
     website,
     studioTitle,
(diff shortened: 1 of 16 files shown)


frontend/src/Movie/Index/Table/MovieIndexRow.tsx
```diff
@@ -162,7 +163,7 @@ function MovieIndexRow(props: MovieIndexRowProps) {
         if (name === 'sortTitle') {
           return (
             <VirtualTableRowCell key={name} className={styles[name]}>
-              <Link to={`/movie/${movieId}`}>{title}</Link>
+              <MovieTitleLink titleSlug={titleSlug} title={title} />
```
plz12345, line 166: @sampulsar This is good, but I noticed you missed the MovieIndexPoster and MovieIndexOverview card.

frontend/src/Studio/Details/SceneRow.tsx
plz12345, line 1: @sampulsar It won't break, but for consistency, can you make these same edits in SceneIndexPoster and SceneIndexOverview ?

src/Whisparr.Api.V3/Movies/MovieController.cs
```diff
@@ -263,8 +263,19 @@ public ActionResult<MovieResource> GetMovieById(string id)
 
             if (!isNumeric)
             {
-                var resourceByForeignId = _moviesService.FindByForeignId(id);
-                resource = resourceByForeignId.ToResource(_configService.AvailabilityDelay, _qualityUpgradableSpecification);
+                if (id.StartsWith("tmdbid:"))
+                {
+                    if (int.TryParse(id.Replace("tmdbid:", ""), out var tmdbid))
```
plz12345: @sampulsar requesting just `tmdb:` as the prefix.

src/Whisparr.Api.V3/Movies/MovieResource.cs
```diff
@@ -115,6 +115,9 @@ public static MovieResource ToResource(this Movie model, int availDelay, IUpgrad
 
             var collection = model.MovieMetadata.Value.CollectionTmdbId > 0 ? new MovieCollectionResource { Title = model.MovieMetadata.Value.CollectionTitle, TmdbId = model.MovieMetadata.Value.CollectionTmdbId } : null;
 
+            // Convert to tmdbid: if TmdbId is present, to allow GetMovieById(string id) to be used.
+            var titleSlug = model.TmdbId > 0 ? $"tmdbid:{model.TmdbId}" : model.MovieMetadata.Value.ForeignId;
```
plz12345: @sampulsar requesting just `tmdb:` as the prefix.

plz12345 (changes requested):

frontend/src/Movie/Index/Table/MovieIndexRow.tsx
sampulsar: @plz12345 Good catch

src/Whisparr.Api.V3/Movies/MovieController.cs
sampulsar: tmdb: is now used for the titleSlug for TMDB items

plz12345 (approved):

PR #134, opened 2026-02-23
Fix: Unable to use "is not" filters after server-side filtering enabled

plz12345:
#### Database Migration
NO

#### Description
The BaseRepository didn't have the use of a "NOT IN" SQL builder type.  This was never an issue because this type of filtering would normally happen client-side, but after moving to server-side filtering, it became apparent.  Simple repro would be a filter like "performer hair color is not blonde" on the front end, which would translate the raw SQL without the `NOT`, and actuyally invert the expected results.

Tests pass and UX passes locally, but I don't have Postgres in dev env, so relying on CI to tell us if that is OK.

#### Todos
- [x] Tests

- Fixes whisparr/whisparr#1019


in1323:
I am the OP of the issue mentioned, I use a postgres database and the PR behaves how I would expect and when merged I would consider the issue resolved

sampulsar (approved):

PR #135, opened 2026-02-23
New: Movie and Scene index pagination, performance improvements

plz12345:
#### Database Migration
YES - 024

#### Description
- Movie and Scene index pagination, page size from 5 to 1000 items
- Revised sorting and filtering for both index pages
- Performance tuned to load movie and scene pages in <100ms, most are <50ms when using SSD
- Moved MediaCoverService to cache-first to avoid disk scans when attempting to map images to local image cache.  It refreshes on app start and on the first cover mapper-dependent API call each 15 minutes.  Was previously happening on every API call.  This was a choke point on very large/huge libraries, especially if config was not on SSD.
- Movie and Scene search performance improved, lookup is now a one-shot API call that will determine existing items compared to metadata results.
- New database indices on `MovieMetadata` for faster sorting and filtering.
- Index footer now displays item counts for scenes (or movies), not just current filtered ones
- Monitor movie/scene toggle performance improved by 10x (240ms -> 25ms in testing)

The app load API call to `/movie/bulk` is now removed.  On a 15K item library, this was usually about 50MB of JSON that the browser had to churn through.  Added a screenshot of a fresh page load on a 26K item library, showing the new normal set of init API calls, which now excludes the bulk ones for movie/scene, as well as previous PR's that removed performer and studio.

#### API-related

All of these will be reflected in SwaggerUI documentation
- `POST /movie/paged?itemType=scene|movie` - Supports index pagination, sorting, and filtering
- `PATCH /movie/{id}` - Replaces `PUT /movie/{id}` for toggling movie/scene monitor
- `POST /movie/list` - Gets a list of movies based on foreignId's
- `GET /movie/lookup?term={term}` - Revised to include whther the movie already exists in the library
- `GET /scene/lookup?term{term}` - Same as /movie/lookup

<details>
image

image

image

</details>

#### Other Notes
- The jumpbar (letters down the right side of page) when sorting by Title) was removed.  This makes less sense when the overall view isn't infinite, and had conflicts with the pagination components.
- Header search isn't working at the moment.  Will be fixed prior to merge.

#### Todos
- [x] Tests

- Refs whisparr/whisparr#961
- Fixes whisparr/whisparr#1033


frontend/src/Components/SignalRConnector.js
```diff
@@ -185,6 +180,16 @@ function updateStudioQueryCache(updatedStudio) {
...
 }
 
+function removeMovieQueryCache(updatedMovie) {
+  if (!updatedMovie || !updatedMovie.foreignId) {
+    return;
+  }
+
+  queryClient.removeQueries({
+    queryKey: [`/performer/${updatedMovie.foreignId}`]
+  });
```
Gykes, line 190: Should this not be removing movie cache?

/movie instead of /performer

frontend/src/Scene/Index/useSceneIndexQuery.ts
```diff
@@ -0,0 +1,61 @@
...
+    // String key indicates a predefined filter
+    filterDef = movieIndexFilters.find((f) => f.key === selectedFilterKey);
+    filters = filterDef && filterDef.filters ? filterDef.filters : [];
+  }
+
+  const queryBody = {
+    ...params,
+    filters: [
+      // Always filter for scenes only
+      { key: 'itemtype', type: 'equal', value: 'scene' },
```
Gykes: I know the backend normalizes but might be a good idea to keep `itemType` universal. 

UseMovieIndexQuery capitilizes the T.

Wont cause issues, just cosmetic.

frontend/src/Movie/useMovie.ts
```diff
@@ -11,77 +11,74 @@ const AUTH_HEADERS = {
...
+      moveFiles?: boolean;
+    }) => {
+      const url = moveFiles
+        ? `/movie/${movie.id}?moveFiles=true`
+        : `/movie/${movie.id}`;
+      return apiPut<Movie, Movie>(url, movie);
+    },
     onSuccess: (data) => {
-      queryClient.setQueryData([`/movie/${idOrForeignId}`], data);
+      queryClient.setQueryData([`/movie/${data.id}`], data);
```
Gykes: ```queryClient.setQueryData([`/movie/${data.id}`], data);```

`useMovie` reads from `/movie/${titleSlug}. data.id` is a numeric ID but `titleSlug` is a string slug (e.g. "some-movie-title"). This `setQueryData` writes to a key that nothing reads, so the detail page won't reflect the saved changes until a manual refetch.

Should it not be: 
```queryClient.setQueryData([`/movie/${data.titleSlug}`], data);```

Looks like line 59 does it the correct way.

File: frontend/src/Movie/Index/Select/Edit/useEditMovieModal.ts [hunk 4]
```diff
@@ -0,0 +1,120 @@
...
+  // Mutation for saving movie
+  const saveMutation = useApiMutation<Movie, Movie>({
+    method: 'PUT',
+    path: `/movie/${movie.id}`,
+    mutationOptions: {
+      onSuccess: (data) => {
+        if (data?.foreignId) {
+          // Invalidate React Query cache for this movie
+          queryClient.invalidateQueries({
+            queryKey: [`/movie/${data.foreignId}`],
```
Gykes: Same issue as `useMovie.ts` 

I think this should be :

```queryClient.invalidateQueries({ queryKey: [`/movie/${data.titleSlug}`] });```

File: src/NzbDrone.Core/Datastore/PagedFilters/NumericFilterInt.cs [hunk 5]
```diff
@@ -0,0 +1,83 @@
...
+            var property = propertySelector.Body;
+
+            switch (operation)
+            {
+                case "equal":
+                    var equalExpr = Expression.Lambda<Func<T, bool>>(
+                        Expression.Call(
+                            typeof(Enumerable),
+                            "Contains",
+                            new[] { typeof(int) },
```
Gykes, line 36: Both the "equal" and "not equal" cases pass `typeof(int)` as the generic type argument to Enumerable.Contains, but values is a `List<long>` (from ParseLongArray) and property is a long expression (line 16). The types don't match `Enumerable.Contains<int>` can't accept a `List<long>` and a long member expression. This will probably throw an error at runtime when a user filters `sizeOnDisk` with equal/notEqual.

frontend/src/Components/SignalRConnector.js
```diff
@@ -387,38 +392,29 @@ class SignalRConnector extends Component {
...
-    const section = 'movies';
-
     // Support batch payloads (Resources) and single (resource)
     if (Array.isArray(body.resources) && body.resources.length > 0) {
       // Batched update
       if (body.action === 'updated') {
-        this.props.dispatchUpdateItemsBatch(body.resources.map((resource) => ({ section, ...resource })));
         body.resources.forEach(updateMovieInPerformerWorksQueryCache);
         body.resources.forEach(updateMovieInStudioWorksQueryCache);
         body.resources.forEach(updateMovieInStudioWorksQueryCache);
```
Gykes: I know this isn't your diff, but duplication?

frontend/src/Movie/Index/Select/Delete/useDeleteMoviesMutation.ts
```diff
@@ -0,0 +1,55 @@
...
+          {
+            method: 'DELETE',
+            headers: {
+              'X-Api-Key': window.Whisparr.apiKey,
+            },
+          }
+        );
+
+        if (!response.ok) {
+          throw new Error(`Failed to delete performer ${id}`);
```
Gykes: Should this error not refer to movies

frontend/src/Scene/Index/Select/Delete/useDeleteScenesMutation.ts
```diff
@@ -0,0 +1,55 @@
...
+          {
+            method: 'DELETE',
+            headers: {
+              'X-Api-Key': window.Whisparr.apiKey,
+            },
+          }
+        );
+
+        if (!response.ok) {
+          throw new Error(`Failed to delete performer ${id}`);
```
Gykes: Should this error not refer to scenes

frontend/src/Scene/Index/Select/SceneIndexSelectFooter.tsx
```diff
@@ -153,45 +148,42 @@ function SceneIndexSelectFooter() {
...
+            onPress={onDeleteSelectedPress}
           >
             {translate('Delete')}
           </SpinnerButton>
         </div>
       </div>
 
       <div className={styles.selected}>
-        {translate('ScenesSelectedInterp', { count: selectedCount })}
+        {translate('MoviesSelectedInterp', { count: selectedCount })}
```
Gykes: Can probably revert this line as it should be: `{translate('ScenesSelectedInterp', { count: selectedCount })}`

frontend/src/Movie/Index/Posters/MovieIndexPosters.tsx
```diff
@@ -125,206 +49,30 @@ export default function MovieIndexPosters(props: MovieIndexPostersProps) {
...
-          },
-          items,
-          sortKey,
-          isSelectMode,
-        }}
-      >
-        {Cell}
-      </Grid>
+      <div style={{ display: 'flex', flexWrap: 'wrap' }}>
+        {items.map((movie) => (
```
Gykes, line 61: This one is a question/concern and not a bug:

So, with the removal of react-window virtualization in leu of a plain map it means that users can render 1k posters at a time. On shittier machines this could potentially cause lots of issues, no? Super laggy display, high memory usage, stuff like that. Would it be better to keep your way for up to say 100-200 scenes but then virtualize the rest to help with hardware managment?

frontend/src/Movie/Index/Select/Delete/useDeleteMovieModalFooterHandler.ts
```diff
@@ -0,0 +1,31 @@
...
+  // Returns a function that takes delete options and performs the deletion
+  const handleDelete = useCallback(
+    (deleteFiles: boolean, addImportExclusion: boolean) => {
+      onModalClose(); // Close modal immediately when delete is confirmed
+      deleteMutation.mutate({
+        movieIds,
+        options: { deleteFiles, addImportExclusion },
+      });
+    },
+    [movieIds, onModalClose, deleteMutation]
```
Gykes: `deleteMutation` is the full object from `useMutation()`. While mutate itself is reference-stable, the object containing it gets a new reference whenever the mutation's status changes. So this `useCallback` re-fires at least once per mutation lifecycle.

This is less of a bug and just a very minor performance thing and more of a code hygiene issue than anything else

src/NzbDrone.Core/MediaCover/MediaCoverService.cs
```diff
@@ -693,10 +707,16 @@ private Dictionary<string, FileInfo> GetCoverFileInfos(string subFolder)
                     .ToDictionary(x => x.FullName, PathEqualityComparer.Instance);
         }
 
+        public void Handle(ApplicationStartedEvent message)
+        {
+            GetMovieCoverFileInfos();
+            GetPerformerCoverFileInfos();
+            GetStudioCoverFileInfos();
+        }
```
Gykes, line 715: This being synchronous and the fact that it blocks application startup untill all 3 directory scans are complete could cause slow downs on large libraries. This is still a net improvement over the old behavior (scanning on every API call), but could be made non-blocking by changing to `HandleAsync(ApplicationStartedEvent)` or keeping the lazy-load approach where the first API call populates the cache.

src/NzbDrone.Core/MediaCover/MediaCoverService.cs
```diff
@@ -693,10 +707,16 @@ private Dictionary<string, FileInfo> GetCoverFileInfos(string subFolder)
...
+
         public void HandleAsync(MovieUpdatedEvent message)
         {
             var updated = EnsureCovers(message.Movie);
```
Gykes, line 719: Not in your diff just I noticed it.

The cache is populated at startup and lazy refreshed. When covers are actually changed in anyway then the cache is never invalidated.  New data wont populate untill after the 15 mins TTL. I think you can fix it by adding a `.clear()` after each mutation.

`_movieCoverFileInfosCache.Clear();`

This also applies to performer and studio handlers on this page

frontend/src/Movie/Index/MovieIndex.tsx
Gykes, line 1: Making a general comment on this one. Looks like The jumpbar was removed. I know this is due to how the new pagination works and the backend would need to resolve where you want to jump but just noting that I think people will notice.

Maybe I am crazy and it got moved and I just dont see it for whatever reason.

frontend/src/Components/SignalRConnector.js
```diff
@@ -387,38 +392,29 @@ class SignalRConnector extends Component {
   };
 
   handleMovie = (body) => {
-    const section = 'movies';
-
     // Support batch payloads (Resources) and single (resource)
     if (Array.isArray(body.resources) && body.resources.length > 0) {
       // Batched update
       if (body.action === 'updated') {
-        this.props.dispatchUpdateItemsBatch(body.resources.map((resource) => ({ section, ...resource })));
```
Gykes, line 396: I presume this was removed as you can't patch individual items into the paginated response. You could possible change the SignalR handler to refetch so that the current page can still auto update

src/Whisparr.Api.V3/Movies/MovieController.cs
```diff
@@ -465,18 +524,99 @@ public ActionResult<MovieResource> UpdateMovie([FromBody] MovieResource moviesRe
             }
 
             var model = moviesResource.ToModel(movie);
-
             var updatedMovie = _moviesService.UpdateMovie(model);
+            return Accepted(MapToResource(updatedMovie));
```
Gykes, line 528: So, this is probably a better way to handle this than the old way. Since you are moving from a bare integar to a full JSON object have you updated meta to handle this? I think this would break the API call, no?

src/NzbDrone.Core/Datastore/PagedFilters/StringFilter.cs
Gykes, line 1: No namespace declaration. All the other filter class in that folder has one. This puts `StringFilter` in the global namespace. It works, but could collide with other types and is inconsistent.

src/NzbDrone.Core/Datastore/PagedFilters/TagFilter.cs
```diff
@@ -0,0 +1,124 @@
...
+            // Build OR chain (contains ANY)
+            Expression BuildOr()
+            {
+                Expression orExpr = null;
+
+                foreach (var id in tagIds)
+                {
+                    var contains = Expression.Call(
+                        tagsExpr,
+                        typeof(IEnumerable<int>).GetMethod("Contains", new[] { typeof(int) }),
```
Gykes, line 71: Contains is a LINQ extension method on `System.Linq.Enumerable`, not an instance method on `IEnumerable<int>`. `GetMethod` on the interface type returns null. This will throw `NullReferenceException` when `Expression.Call` tries to use it.

src/NzbDrone.Core/Datastore/PagedFilters/TagFilter.cs
```diff
@@ -0,0 +1,124 @@
...
+            // Build AND chain (contains ALL)
+            Expression BuildAnd()
+            {
+                Expression andExpr = null;
+
+                foreach (var id in tagIds)
+                {
+                    var contains = Expression.Call(
+                        tagsExpr,
+                        typeof(IEnumerable<int>).GetMethod("Contains", new[] { typeof(int) }),
```
Gykes, line 89: Same as line 71

src/Whisparr.Api.V3/Movies/Helpers/Paging.cs
```diff
@@ -0,0 +1,129 @@
...
+                    case "path": // TODO: test
+                        StringFilter.Apply(pageSpec, jsonElement, op, p => p.Path);
+
+                        break;
+                    case "qualityprofileid": // TODO: test
+                        NumericFilterInt.Apply(pageSpec, jsonElement, op, p => p.QualityProfileId);
+
+                        break;
+                    case "releasedate": // OK
+                        DateFilter.Apply(pageSpec, jsonElement, op, p => (DateTime)p.MovieMetadata.Value.ReleaseDateUtc);
```
Gykes: `ReleaseDateUtc` is `DateTime?` (nullable) per `MovieMetadata.cs:34`. The hard cast (DateTime) will throw `InvalidOperationException` at runtime for any movie that has a null release date. This expression is compiled into a filter applied to every row matching the query. A single movie with no release date will crash the entire paged request when the user sorts/filters by release date. I know release date is a required field in meta so I doubt this would cause issues but might be smart to make this null save. 

`p => p.MovieMetadata.Value.ReleaseDateUtc ?? DateTime.MinValue`

frontend/src/Movie/Index/Select/Edit/useEditMovieModal.ts
```diff
@@ -0,0 +1,120 @@
...
+  const isPathChanging =
+    pendingChanges.rootFolderPath != null &&
+    movie.rootFolderPath !== pendingChanges.rootFolderPath;
+
+  // Build settings
+  const movieSettings: MovieSettings = {
+    monitored: movie.monitored,
+    qualityProfileId: movie.qualityProfileId,
+    minimumAvailability: (
+      movie as unknown as {
```
Gykes: `overview` (line 109) and `path` (line 113) are already defined on the Movie interface, the as unknown as casts are unnecessary. For `minimumAvailability` and `searchOnAdd`, add them to the Movie type in `Movie.ts` instead of double-casting. This avoids bypassing type safety entirely.

Gykes (dismissed):
Static review only, no actual testing. 

Not 100% familiar with your codebase so things may be a smidge different. Feel free to ignore anything that is dumb

frontend/src/Components/SignalRConnector.js
plz12345: good catch, I'd copied from Performer

frontend/src/Components/SignalRConnector.js
plz12345: fixed

frontend/src/Movie/Index/MovieIndex.tsx
plz12345: You're not crazy, I will add it to the PR description.  It is indeed due to how the pagination works, I had trouble using it inline with TablePager, so I gutted it.

frontend/src/Movie/Index/Posters/MovieIndexPosters.tsx
plz12345: While I agree in principle, if someone is dumb enough to try loading 1000 posters on a low end machine, they deserve the one-time lag.  I say one-time, because they will either complain and we'll tell them to make the pageSize smaller, or, they will realize it themselves.

Also keep in mind that the images are still lazy load, so the hit is reallt DOM, not network, as you note.  I went for simpler, since there is a ceiling to the item count.

frontend/src/Movie/Index/Select/Delete/useDeleteMovieModalFooterHandler.ts
plz12345: I understood some of those words.  Suggestion?

frontend/src/Movie/Index/Select/Delete/useDeleteMoviesMutation.ts
plz12345: fixed

frontend/src/Movie/useMovie.ts
plz12345: Good catch, fixed.

frontend/src/Movie/Index/Select/Delete/useDeleteMovieModalFooterHandler.ts
Gykes: The `useCallback` dep array has `deleteMutation`, which is the whole object. React sees this as new everytime the mutations runs. It also means that `handleDelete` get recreated when it doesnt need to be. 

```
  export function useDeleteMovieModalFooterHandler({
    movieIds,
    onModalClose,
  }: UseDeleteMovieModalFooterHandlerProps) {
    const { mutate, isPending } = useDeleteMoviesMutation();

    const handleDelete = useCallback(
      (deleteFiles: boolean, addImportExclusion: boolean) => {
        onModalClose();
        mutate({
          movieIds,
          options: { deleteFiles, addImportExclusion },
        });
      },
      [movieIds, onModalClose, mutate]
    );

    return {
      onDeletePress: handleDelete,
      isPending,
    };
  }
```

You could do something like this but verify me before blindly trusting it

src/Whisparr.Api.V3/Movies/Helpers/Paging.cs
plz12345: Fixed, good catch

src/NzbDrone.Core/MediaCover/MediaCoverService.cs
plz12345: That is on purpose to avoid disk thrashing during batch processing (which is all serial).  When you batch add 100 movies, you don't want this nuking the cache 100 times, and this is opportunistic anyways, images are going to load one way or another.

frontend/src/Scene/Index/Select/Delete/useDeleteScenesMutation.ts
plz12345: fixed

frontend/src/Scene/Index/useSceneIndexQuery.ts
plz12345: fixed

frontend/src/Components/SignalRConnector.js
plz12345: Nope, it was removed because it's for Redux, which is no longer applicable.  React Query updates happen instead.

frontend/src/Scene/Index/Select/SceneIndexSelectFooter.tsx
plz12345: fixed, copy/paste fun

src/NzbDrone.Core/Datastore/PagedFilters/StringFilter.cs
plz12345: Good catch, fixed

src/Whisparr.Api.V3/Movies/MovieController.cs
plz12345: I'm just using a different overload.  Accepted(movie.id) would just do (another) lookiup and return that object.  Since I already have it, here, this is actually (barely, slightly) faster.  Has nothing to do with meta at all, this is local client updating a movie and local client getting a response body back.  In the edit modal and formerly on the monitor toggle button, it did nothing with the responses anyways.The reason I did this was because MapToResource also gets the statistics populated, which the bare Accpted overload didn't do.

src/NzbDrone.Core/MediaCover/MediaCoverService.cs
plz12345: Yeah, it's a race condition already.  On one hand, having this be primed before page load is good because that first page load is going to be the scene index, which will likely rely on these.  And if this is async, that load could call this same logic if the processing is slow enough to not finish before the page load fires and triggers the /movie/paged API call.  I figured it'd be better to block init to avoid that race.  FWIW, this isn't super slow, it's just, not <100ms fast>.  You'd need 100k+ to even notice it, and even then, we're talking <10 seconds likely.  It's not scanning each and every folder one by one.

My testing was on 26K, and the realtime covermapper was 1-2 seconds.  Unacceptable for every API call, but totally fine for app init.

src/NzbDrone.Core/Datastore/PagedFilters/TagFilter.cs
plz12345: it's protected from that by the foreach loop.  No id in ids = no call to contains.  Keep me honest.

src/NzbDrone.Core/Datastore/PagedFilters/NumericFilterInt.cs
plz12345: I think you are mixing this with NumberFilterLong.  They are separate for this exact reason.  That said, the NumberFilterLong had this bug, and I fixed it in that class.

src/NzbDrone.Core/Datastore/PagedFilters/TagFilter.cs
plz12345: Also same as other comment.  :)

frontend/src/Movie/Index/Select/Edit/useEditMovieModal.ts
plz12345: You are right.  The behavior would be fine for scenes, but not movies.  Fixed.

frontend/src/Movie/Index/Select/Edit/useEditMovieModal.ts
plz12345: these aren't even used, so it was even worse.  Removed.

frontend/src/Movie/Index/Select/Delete/useDeleteMovieModalFooterHandler.ts
plz12345: I'm not fully following your feedback, but it doesn't break, so, Ok.

frontend/src/Movie/Index/Select/Edit/useEditMovieModal.ts
plz12345: It was worse, these aren't used or needed on this modal, anyways.  Artifacts leftover from cloning logic form the AddNewMovie modal, I think.  Cleaned those out.

sampulsar (commented):
I think there is a regression on the do es not contain test PR 134

plz12345:
@sampulsar 

Not sure what you mean.  The changes in both WhereBuilder classes as well as the new unit tests are all there.

Examples:
https://github.com/plz12345/Whisparr-Eros/blob/movie-pagination/src/NzbDrone.Core/Datastore/WhereBuilderSqlite.cs#L60
https://github.com/plz12345/Whisparr-Eros/blob/movie-pagination/src/NzbDrone.Core/Datastore/WhereBuilderPostgres.cs#L59-L60

sampulsar:
@plz12345 

Setup a not contains test for that PR, but it is now behaving as the contains again
image
I added the task to a single item, the first item in the library, and that is the only one that shows.

Changing between contains and not contains shows the same item

plz12345:
You're right, though the unit test is still passing.  Weird, but will fix.

sampulsar (approved):
There is a minor issue but requires a lot of work to correct, and found it on the other paging

plz12345:
@sampulsar I already have a fix for the tag filtering coming.  The filtering I built does need to be re-applied to cover the Performer and Studio index pages, but that will definitely be a separate PR.  Let me get this commit sorted out, before we merge to develop.

sampulsar (approved):

PR #141, opened 2026-02-27
Fix: MovieEditedEvent causes large scan of tracked downloads

plz12345:
#### Database Migration

NO

#### Description

I was troubleshooting significant latency in editing movie/scene items, which appeared as 2500ms when toggling the movie monitor.

I found the issue in a synchronous event handler in TrackedDownloadService, intended to update tracked items when a library item is edited.

The problem was a comparison of “event movie TMDB ID = Tracked Download TMDB ID). Since all StashDB scenes have a default TMDB ID of 0, as do all un-matched downloads, this comparison always returns true.

I added an explicit check and moved both single and bulk events to async, as their downstream behavior relies on SignalR and doesn’t require synchronization.

```
2026-02-26 22:14:46.6|Trace|EventAggregator|Publishing MovieEditedEvent
2026-02-26 22:14:46.6|Trace|EventAggregator|MovieEditedEvent -> TrackedDownloadService
... a bunch of parser matching on my active tracked downloads (in Whisparr's category)
2026-02-26 22:14:49.1|Trace|EventAggregator|MovieEditedEvent <- TrackedDownloadService

```
~2+ seconds of toggle monitor lag, finally explained.

#### Todos

- [x] Tests

### DIFF
--- a/src/NzbDrone.Core/Download/TrackedDownloads/TrackedDownloadService.cs
+++ b/src/NzbDrone.Core/Download/TrackedDownloads/TrackedDownloadService.cs
@@ -31,7 +31,7 @@ namespace NzbDrone.Core.Download.TrackedDownloads
                                           IHandle<MovieAddedEvent>,
                                           IHandleAsync<MovieEditedEvent>,
                                           IHandleAsync<MoviesBulkEditedEvent>,
-                                          IHandle<MoviesDeletedEvent>
+                                          IHandleAsync<MoviesDeletedEvent>
     {
         private readonly IParsingService _parsingService;
         private readonly IHistoryService _historyService;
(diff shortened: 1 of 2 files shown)


src/NzbDrone.Core/Download/TrackedDownloads/TrackedDownloadService.cs
```diff
@@ -280,11 +280,12 @@ public void Handle(MovieAddedEvent message)
             }
         }
 
-        public void Handle(MovieEditedEvent message)
+        public void HandleAsync(MovieEditedEvent message)
         {
             var cachedItems = _cache.Values
                 .Where(t =>
                     t.RemoteMovie?.Movie != null &&
+                    message.Movie.TmdbId != 0 &&
```
sampulsar: Is this correct, as it will only work for TMDB movies. The list will be empty

plz12345: You are right.  I did this while tired.  I've re-worked that logic into a helper method since it was repeated in 3 different events, and also moved all these events to async since none rely synchronous action.  I also added null safety on everything.

sampulsar (approved):

PR #142, opened 2026-02-28
Fix: Hide Movie Monitor if it can not be used

sampulsar:
#### Database Migration

NO

#### Description

The details page for performer and studio will show the movie monitor if it can be enabled. Now the edit modal also has the same code to stop it from being toggled on and the error not bring visible to the user.

Hide the movie menu if the meta source is set to none

<details>
image

image

</details>

#### Todos

- [x] Tests

- Fixes whisparr/whisparr#1031


frontend/src/Components/Page/Sidebar/PageSidebar.tsx
```diff
@@ -276,6 +277,18 @@ function PageSidebar({ isSidebarVisible, isSmallScreen }: PageSidebarProps) {
     height: `${window.innerHeight - HEADER_HEIGHT}px`,
   });
```
plz12345, line 279: @sampulsar two things.  

Need to wrap this in useMemo.  The behavior at present will hide it if you set it to none, but won't unhide it if you revert.  Also, this can be simplified using `.filter`

```suggestion
const links = useMemo(() => {
  if (generalSettings.whisparrMovieMetadataSource?.toLowerCase() === 'none') {
    return LINKS.filter((link) => link.iconName !== icons.FILM);
  }
  return LINKS;
}, [generalSettings.whisparrMovieMetadataSource]);
```

frontend/src/Performer/Details/PerformerDetails.tsx
```diff
@@ -214,11 +214,12 @@ function PerformerDetails() {
     year,
     movies: movies.filter((m) => m.year === year),
   }));
-  const showMovieMonitorToggle = () => {
+  const showMovieMonitorToggle = (): boolean => {
```
plz12345: @sampulsar I think this should be revised to conditionally display based on both having a non-null ID and also when that ID is from the desired metadata source.  It's misleading otherwise.

frontend/src/Performer/Index/Posters/PerformerIndexPoster.tsx
```diff
@@ -60,6 +61,21 @@ function PerformerIndexPoster(props: PerformerIndexPosterProps) {
     height: `${posterHeight}px`,
   };
 
+  const generalSettings = useGeneralSettings();
+
+  const showMovieMonitorToggle = (): boolean => {
```
plz12345: @sampulsar I'd suggest moving this to a exported utility function, since it will be needed here, on the deatils page, in th eedit modal, and also on cast posters.  Easier to define it once and reuse it everywhere.  That way you also don't need to import the general settings selector in a bunch of places.

frontend/src/Studio/Edit/useEditStudioModal.ts
```diff
@@ -88,6 +90,21 @@ export default function useEditStudioModal(
 
   const settings = selectSettings(studioSettings, pendingChanges, saveError);
 
+  const generalSettings = useGeneralSettings();
```
plz12345, line 93: @sampulsar same as comment on the performer hook.  Move this logic to a shared function so it isn't copy-pasted in 5-10 different places.

plz12345:
Also, need to add this logic to the cast poster implementation of the toggle (when viewing scene/movie details.

plz12345 (changes requested):

sampulsar:
@plz12345 

I am going to abandon this PR as I am really limited on time due to life commitments. 

If you want to do something similar, but it may be better to show an error for movie monitor that it can not be checked instead of hiding it, incase the metadata is removed for a studio or performer, to that they change meta providers and can not save an item.

plz12345:
I can just pick this one up and run with the changes, you had a solid start to it.

plz12345:
I created a new PR with the finished changes, closing this one.

PR #153, opened 2026-03-02
Fix: Hide movies and monitor toggle when unusable/disabled

plz12345:
## Summary
- Hide the movie monitor toggle button when it cannot be used (e.g. no valid monitor configuration)
- Hide the Movie Menu when Metasource is set to None
- Refactored show-toggle logic into a dedicated hook for cleaner component code

partially authored by @sampulsar 

- Fixes whisparr/whisparr#1031


sampulsar (approved):

sampulsar:
@plz12345 

Link to 1031.

There is only one rare issue with hiding the checkbox if someone checks on movie monitor and either the link is removed, eg wrong performer in StashDB, ot they change the movie metasource. They may not be able to add a tag as the validation could fail.

Option would be to display an error it movie monitor can not be checked.

plz12345:
Ack, on the edge case.

PR #202, opened 2026-03-12
Fix: Improve the performance when adding Performers, Studios

sampulsar:
#### Database Migration

NO

#### Description

Alter the Adding of new performers, so it is faster when no performers need to be added.

Check if the performer/studio exists before getting the data from skyhook.

The changes are to stop any processes that may take a long time 100ms, eg large database calls, Skyhook calls, when they are not required. 

<details>

</details>

#### Todos

- [x] Tests

- Fixes whisparr/whisparr#XXXX

### DIFF
--- a/src/NzbDrone.Core/Movies/Performers/AddPerformerService.cs
+++ b/src/NzbDrone.Core/Movies/Performers/AddPerformerService.cs
@@ -68,6 +68,18 @@ namespace NzbDrone.Core.Movies.Performers
             {
                 try
                 {
+                    if (existingPerformerForeignIds.Any(f => f == m.ForeignId))
+                    {
+                        _logger.Debug("Foreign ID {0} was not added due to validation failure: Performer already exists in database", m.ForeignId);
+                        continue;
+                    }
+
+                    if (performersToAdd.Any(f => f.ForeignId == m.ForeignId))
+                    {
+                        _logger.Debug("Foreign ID {0} was not added due to validation failure: Performer already exists on list", m.ForeignId);
+                        continue;
+                    }
+
                     var performer = AddSkyhookData(m);
                     performer = SetPropertiesAndValidate(performer);
 
--- a/src/NzbDrone.Core/Movies/Performers/AddPerformerService.cs
+++ b/src/NzbDrone.Core/Movies/Performers/AddPerformerService.cs
@@ -62,13 +62,13 @@ namespace NzbDrone.Core.Movies.Performers
         {
             var added = DateTime.UtcNow;
             var performersToAdd = new List<Performer>();
-            var existingPerformerForeignIds = _performerService.AllPerformerForeignIds();
+            var existingPerformerForeignIds = new HashSet<string>(_performerService.AllPerformerForeignIds());
 
             foreach (var m in newPerformers)
             {
                 try
                 {
-                    if (existingPerformerForeignIds.Any(f => f == m.ForeignId))
+                    if (existingPerformerForeignIds.Contains(m.ForeignId))
                     {
                         _logger.Debug("Foreign ID {0} was not added due to validation failure: Performer already exists in database", m.ForeignId);
                         continue;
@@ -84,18 +84,6 @@ namespace NzbDrone.Core.Movies.Performers
                     performer = SetPropertiesAndValidate(performer);
 
                     performer.Added = added;
-
-                    if (existingPerformerForeignIds.Any(f => f == performer.ForeignId))
-                    {
-                        _logger.Debug("Foreign ID {0} was not added due to validation failure: Performer already exists in database", m.ForeignId);
-                        continue;
-                    }
-
-                    if (performersToAdd.Any(f => f.ForeignId == performer.ForeignId))
-                    {
-                        _logger.Debug("Foreign ID {0} was not added due to validation failure: Performer already exists on list", m.ForeignId);
-                        continue;
-                    }
 
                     _logger.Info("Adding Performer {0}", performer.Name);
 
--- a/src/NzbDrone.Core/Movies/Performers/PerformerService.cs
+++ b/src/NzbDrone.Core/Movies/Performers/PerformerService.cs
@@ -67,6 +67,11 @@ namespace NzbDrone.Core.Movies.Performers
 
         public List<Performer> AddPerformers(List<Performer> performers)
         {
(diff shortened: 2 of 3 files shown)


src/NzbDrone.Core/Movies/Performers/AddPerformerService.cs
```diff
@@ -68,6 +68,18 @@ public List<Performer> AddPerformers(List<Performer> newPerformers, bool ignoreE
...
+                        _logger.Debug("Foreign ID {0} was not added due to validation failure: Performer already exists in database", m.ForeignId);
+                        continue;
+                    }
+
+                    if (performersToAdd.Any(f => f.ForeignId == m.ForeignId))
+                    {
+                        _logger.Debug("Foreign ID {0} was not added due to validation failure: Performer already exists on list", m.ForeignId);
+                        continue;
+                    }
+
```
plz12345: @sampulsar if the intent is to improve performance, why was this pair of checks duplicated?  It runs both before and after the AddSkyhookData method call.

Additionally, doing a comparision using .Any() and the List iterator is not exectly a performance optimization at scale.  This should be revised to use a HashSet<string> instead of List<string>.  If you are adding a single performer on a DB with 5000, that's 5000 potential lookup iterations it does, where a HashSet is a single lookup.

```suggestion
var existingPerformerForeignIds = new HashSet<string>(_performerService.AllPerformerForeignIds());

if (existingPerformerForeignIds.Contains(m.ForeignId))

```

src/NzbDrone.Core/Movies/Performers/PerformerService.cs
```diff
@@ -67,6 +67,11 @@ public Performer AddPerformer(Performer newPerformer)
 
         public List<Performer> AddPerformers(List<Performer> performers)
         {
+            if (performers.Count == 0)
```
plz12345: @sampulsar this section should have the same optimization, so the comparison logic for foreignId's already in the DB is optimized via HashSet<string> to match existing/append the "to add" list.

src/NzbDrone.Core/Movies/Studios/AddStudioService.cs
```diff
@@ -77,6 +77,18 @@ public List<Studio> AddStudios(List<Studio> newStudios, bool ignoreErrors = fals
 
                 try
                 {
+                    if (existingStudioForeignIds.Any(f => f == m.ForeignId))
```
plz12345: @sampulsar
This should shift to using the HashSet lookup for faster performance.

```suggestion
var existingStudioForeignIds = new HashSet<string>(_studioService.AllStudioForeignIds());

if (existingStudioForeignIds.Contains(m.ForeignId))

**Review by plz12345 — CHANGES_REQUESTED** [review]
Reviewed, 1 structural change, and 2 (higher) performance recommendations made.

File: src/NzbDrone.Core/Movies/Performers/PerformerService.cs (hunk 2 above)
Comment (sampulsar) [reply src/NzbDrone.Core/Movies/Performers/PerformerService.cs]: If AddPerformer does not have any performers passed then it can just return, as there is no need to lookup all the performers.

File: src/NzbDrone.Core/Movies/Performers/PerformerService.cs [hunk 4]
```diff
@@ -67,6 +67,11 @@ public Performer AddPerformer(Performer newPerformer)
 
         public List<Performer> AddPerformers(List<Performer> performers)
         {
+            if (!performers.Any())
+            {
+                return performers;
+            }
+
             var allPerformers = _performerRepo.All();
```
plz12345: My previous comment was more toward the code after your addition to this method.  I understand the intent of having an early exit, but the below uses the same inefficient lookup code against the entire Performers table in memory.  We don't want to need to iterate the entire performers table of data (using .Any and a comparison) when we are just matching against ForeignId and parsing the performers into buckets of existing and to-be-added.  This scales better in larger libraries (thousands to performers already in DB).

```suggestion
            var allPerformers = _performerRepo.All();

            var allPerformerForeignIds = new HashSet<string>(allPerformers.Select(p => p.ForeignId));
            var existing = performers.Where(p => allPerformerForeignIds.Contains(p.ForeignId)).ToList();
            var performersToAdd = performers.Where(p => !allPerformerForeignIds.Contains(p.ForeignId)).ToList();
```

Edit: Side note, maybe _performerService.AllPerformerForeignIds() should just return a `HashSet<string>` in the first place?

src/NzbDrone.Core/Movies/Studios/AddStudioService.cs
```diff
@@ -69,31 +69,31 @@ public List<Studio> AddStudios(List<Studio> newStudios, bool ignoreErrors = fals
...
-
-                    if (existingStudioForeignIds.Any(f => f == studio.ForeignId))
+                    if (existingStudioForeignIds.Contains(m.ForeignId))
                     {
                         _logger.Debug("Foreign ID {0} was not added due to validation failure: Studio already exists in database", m.ForeignId);
                         continue;
                     }
 
-                    if (studiosToAdd.Any(f => f.ForeignId == studio.ForeignId))
+                    if (studiosToAdd.Any(f => f.ForeignId == m.ForeignId))
```
plz12345, line 86: @sampulsar not sure what the intent is on this line.  If the intent is to only hit Skyhook once per Studio being added, this technically does work. If so, I'm good with it as a steam valve to avoid Skyhook repetitions, though de-duplicating `newStudios` is probably better than this.  If the list is already unique, this comparison isn't needed at all.

plz12345 (changes requested):

src/NzbDrone.Core/Movies/Studios/AddStudioService.cs
sampulsar: That is what the code would do as studiosToAdd is a list of studios data that will actually be added after getting the data and verifying it.

It likely never gets hit as the scene movie data should never have a dup. Likely just a safe guard when it was originally added. The max is the list may be triple digits , but usually a single digit

src/NzbDrone.Core/Movies/Performers/PerformerService.cs
sampulsar: Using a HashSet

plz12345 (approved):

PR #212, opened 2026-03-13
New: Refresh the Movies by oldest to newest info sync

sampulsar:
#### Database Migration

NO

#### Description

Alter Refresh movie so that it will start where it left off like Studio and Performers.

The manual refresh will start at the first id every time meaning that the same items are refreshed.

Change to read by LastInfoSync

<details>

</details>

#### Todos

- [ ] Tests

- Fixes whisparr/whisparr#XXXX

### DIFF
--- a/src/NzbDrone.Core/Movies/MovieRepository.cs
+++ b/src/NzbDrone.Core/Movies/MovieRepository.cs
@@ -36,6 +36,7 @@ namespace NzbDrone.Core.Movies
         Movie FindByPath(string path);
         Dictionary<int, string> AllMoviePaths();
         List<int> AllMovieIds();
+        List<int> AllMovieIdsOrderByLastInfoSync();
         List<int> AllMovieTmdbIds();
         List<string> AllMovieTpdbIds();
         List<string> AllMovieStashIds();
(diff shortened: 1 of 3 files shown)


plz12345 (changes requested):
Can you add a new repo method to fetch the Id's sorted by LastInfoSync instead?  That chunking was there for memory optimization, since fetching the entire Movies table is a heavy hit we can fairly easily avoid.

In MovieRepository
```csharp
public List<int> AllMovieIdsOrderByLastInfoSync()
{
    using (var conn = _database.OpenConnection())
    {
        return conn.Query<int>("SELECT \"Movies\".\"Id\" FROM \"Movies\" JOIN \"MovieMetadata\" ON \"Movies\".\"MovieMetadataId\" = \"MovieMetadata\".\"Id\" ORDER BY \"MovieMetadata\".\"LastInfoSync\" ASC").ToList();
    }
}
```

In MovieService
```csharp
public List<int> AllMovieIdsOrderByLastInfoSync()
{
    return _movieRepository.AllMovieIdsOrderByLastInfoSync();
}
```

Then in this method, just swap out the original service method call for the new one, retaining the chunking that you removed in the PR.

Napkin math on the memory usage, though this can vary wildly based on the Description field.
Library size | Full load (GetAllMovies) | Chunked (1000 at a time) | Savings
-- | -- | -- | --
10,000 | ~32 MB | ~3.2 MB | ~29 MB
26,000 | ~83 MB | ~3.2 MB | ~80 MB
100,000 | ~320 MB | ~3.2 MB | ~317 MB

sampulsar:
I will try that again, but when I initially tried the ids where ordered by the id number, so it would do the same thing.

sampulsar:
@plz12345 

Corrected, as I initially tried that style of query, but did not check the data in the database as it looked like it was returning by  id as the lower ones had the oldest date.

A manual change to the date, showed that it actually works.

plz12345 (approved):

plz12345:
Looks good now.  I submitted a similar PR to do the same memory optimization on Performer and Studio, though improvements for those will be smaller since they house less metadata.

PR #239, opened 2026-03-21
Add posters view to PerformerDetails

sk8ersteve:
#### Database Migration

NO

#### Description

Adds a new "Posters" view option to the Performer Details page, allowing users to toggle  between the existing table view and a poster grid view of a performer's scenes/works. Includes   a view menu in the toolbar, configurable poster sizes (small/medium/large) via an options  modal, and responsive grid layout. Each poster card shows the scene image, title, year,  runtime, and a progress bar.

<details>

Looks the same as the scenes page but only shows scenes for one performer.

</details>

#### Todos

- [ ] Tests

- Fixes whisparr/whisparr#284


plz12345:
Dropping a 450+ line commit with no iteration history, no screenshots for what is clearly a UI-centric change, and no indicator you actually tested it (followed our template)?  As your first action since creating an account, no less...

I'm going to trigger the build since it's UI-only, but please come back with a SFW-enabled screenshot of the UI changes.

plz12345:
Also
> Each poster card shows the scene image, title, year, runtime, and a progress bar

That is not reflected in the code you pushed, which I suspect is un-tested LLM-generated code based on a few indicators in this PR.

sk8ersteve:
Sorry yes. Full transparency, yes it was LLM generated, but it was tested locally and does solve my exact problem. I created a fresh account for this because my main account is associated with my job and yeah I don't want to mix that up with this (I also had to correct the email that auto populated in the commit hence the new commit timestamp). Hopefully that's understandable. I made this change like a month ago with Claude code but didn't have the time to properly dig through and understand it so I put it off, and then I grew impatient with not having the feature on my actual server so I decided to just try to submit the PR and hope for the best while not even remembering how exactly I got it to build and run locally. 

When I get a chance, I'll try again to get some pictures for you and maybe better document it. Everything did work locally running it on my laptop with a few videos downloaded from my server where I'm still only running the official build. I'll understand if you don't want to touch this until I put some more work into it.

Edit: if you're curious about my github history, my main account posted a comment on the issue I tagged as this fixing. Similar username.

sk8ersteve:
Also I didn't check off "Tests" because I wasn't sure if testing it locally satisfies that requirement. In my day-to-day work, we do test suites for every code change, and I haven't dug into what formal testing stuff you guys have.

sk8ersteve:
Ok I actually did figure out running locally. Tell me, is using the built-in hide feature that blurs the thumbnails good enough, or do I really need to get some different thumbnails in here?

plz12345:
I guess my biggest concern is the duplication of components, rather than reusing existing ones.  Like the posters/poster component.  There shouldn't have been a need to duplicate those for this.  Also, need to see how this actually looks, both with scenes, and movies, since a performer can have both.

sk8ersteve:
Ok here's my second attempt, reusing existing components rather than duplicating: https://github.com/Whisparr/Whisparr-Eros/pull/252

I did have that same concern upon my first reading of the code but for some reason I thought it made sense at the time to create the new component.

PR #248, opened 2026-03-26
Fix: Additional sorting of scenes

sampulsar:
#### Database Migration

NO

#### Description

Additional sorting as the selected sort may not be unique enough, eg studio. 

<details>

</details>

#### Todos

- [x] Tests

- Fixes whisparr/whisparr#1069

### DIFF
--- a/src/NzbDrone.Core.Test/Datastore/BasicRepositoryFixture.cs
+++ b/src/NzbDrone.Core.Test/Datastore/BasicRepositoryFixture.cs
@@ -28,6 +28,7 @@ namespace NzbDrone.Core.Test.Datastore
                 .CreateListOfSize(5)
                 .All()
                 .With(x => x.Id = 0)
+                .With(x => x.Interval = 1)
                 .BuildList();
         }
 
@@ -323,6 +324,22 @@ namespace NzbDrone.Core.Test.Datastore
             data.PageSize.Should().Be(2);
             data.TotalRecords.Should().Be(_basicList.Count);
             data.Records.Should().BeEquivalentTo(_basicList.OrderByDescending(x => x.Id).Skip((page - 1) * 2).Take(2));
+        }
+
+        [TestCase(1, 2)]
+        [TestCase(2, 2)]
+        [TestCase(3, 1)]
+        public void get_paged_should_work_with_default_sort_keys(int page, int count)
+        {
+            Subject.InsertMany(_basicList);
+            var pageSpec = new PagingSpec<ScheduledTask>() { Page = page, PageSize = 2, SortKey = "Interval", SortDirection = SortDirection.Descending, DefaultSortKeys = new List<string> { "LastExecution" } };
+            var data = Subject.GetPaged(pageSpec);
+
+            data.Page.Should().Be(page);
+            data.PageSize.Should().Be(2);
+            data.TotalRecords.Should().Be(_basicList.Count);
+            data.Records.Should().BeEquivalentTo(_basicList.OrderBy(x => x.LastExecution).OrderByDescending(x => x.Interval).Skip((page - 1) * 2).Take(2));
+            data.Records.Should().NotBeEquivalentTo(_basicList.OrderByDescending(x => x.LastExecution).OrderByDescending(x => x.Interval).Skip((page - 1) * 2).Take(2));
         }
     }
 }
--- a/src/NzbDrone.Core/Datastore/BasicRepository.cs
+++ b/src/NzbDrone.Core/Datastore/BasicRepository.cs
@@ -452,13 +452,25 @@ namespace NzbDrone.Core.Datastore
                 pagingSpec.SortKey = $"{_table}.{_keyProperty.Name}";
             }
 
+            var orderByClause = BuildOrderByClause(pagingSpec);
+            builder.OrderBy(orderByClause);
+
+            return queryFunc(builder).ToList();
+        }
+
+        private string BuildOrderByClause(PagingSpec<TModel> pagingSpec)
+        {
+            var pagingOffset = Math.Max(pagingSpec.Page - 1, 0) * pagingSpec.PageSize;
+
             var sortKey = TableMapping.Mapper.GetSortKey(pagingSpec.SortKey);
             var sortDirection = pagingSpec.SortDirection == SortDirection.Descending ? "DESC" : "ASC";
-            var pagingOffset = Math.Max(pagingSpec.Page - 1, 0) * pagingSpec.PageSize;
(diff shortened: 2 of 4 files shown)


src/NzbDrone.Core/Datastore/BasicRepository.cs
```diff
@@ -455,7 +455,23 @@ protected List<TModel> GetPagedRecords(SqlBuilder builder, PagingSpec<TModel> pa
...
+                    {
+                        continue;
+                    }
+
+                    orderByClause += $", \"{key.Table ?? _table}\".\"{key.Column}\" ASC";
+                }
+            }
+
+            orderByClause += $" LIMIT {pagingSpec.PageSize} OFFSET {pagingOffset}";
+            builder.OrderBy(orderByClause);
```
plz12345: @sampulsar can you extend the unit tests to cover this new logic?  This is very low level, so any change here is very high risk.

src/NzbDrone.Core/Datastore/PagingSpec.cs
```diff
@@ -9,6 +9,7 @@ public class PagingSpec<TModel>
         public int Page { get; set; }
         public int PageSize { get; set; }
         public int TotalRecords { get; set; }
+        public string DefaultSortKeys { get; set; }
```
plz12345:
```suggestion
        public List<string> DefaultSortKeys { get; set; }
```

plz12345: @sampulsar, suggesting a List<string> here, to avoid the use of string split in low-level DB code.  It's inherently more fragile than it needs to be.

src/Whisparr.Api.V3/Movies/MovieController.cs
```diff
@@ -897,6 +897,7 @@ private ActionResult<PagingResource<MovieResource>> GetPagedMoviesWithTags(Movie
         /// <summary> Standard paged movie index without tag filtering. </summary>
         private ActionResult<PagingResource<MovieResource>> GetPagedMoviesStandard(MoviePagingRequestResource request, NzbDrone.Core.Datastore.PagingSpec<Movie> pageSpec)
         {
+            pageSpec.DefaultSortKeys = "movieMetadata.ReleaseDate,movieMetadata.StudioTitle,movieMetadata.SortTitle";
```
plz12345:
```suggestion
            pageSpec.DefaultSortKeys = new List<string> { "movieMetadata.ReleaseDate", "movieMetadata.StudioTitle", "movieMetadata.SortTitle" };

```

src/NzbDrone.Core/Datastore/BasicRepository.cs
```diff
@@ -455,7 +455,23 @@ protected List<TModel> GetPagedRecords(SqlBuilder builder, PagingSpec<TModel> pa
...
+            {
+                foreach (var defaultSortKey in pagingSpec.DefaultSortKeys.Split(","))
+                {
+                    var key = TableMapping.Mapper.GetSortKey(defaultSortKey);
+                    if (key.Column == sortKey.Column && key.Table == sortKey.Table)
+                    {
+                        continue;
+                    }
+
+                    orderByClause += $", \"{key.Table ?? _table}\".\"{key.Column}\" ASC";
```
plz12345: @sampulsar Sonar suggests using a stringbuilder for this, vs. the +=, as it's in a loop.  I know it's not high-volume logic, but it would be idiomatic to use that here.

You could even break that into a private static helper method to make that specific logic easier to unit test outside of the actual filtering.

plz12345 (changes requested):

src/NzbDrone.Core/Datastore/PagingSpec.cs
sampulsar: List used

src/NzbDrone.Core/Datastore/BasicRepository.cs
sampulsar: Stringbuilder used

src/NzbDrone.Core/Datastore/BasicRepository.cs
sampulsar: Add get_paged_should_work_with_default_sort_keys unittest

sampulsar:
@plz12345 
unit test added to cover when the default sort keys are set, as the Intervals are all set the same the default sort leys will then do a secondary sort to make the order unique.

The existing unit tests cover when it is not set.

sampulsar:
@plz12345 ready for another review

src/Whisparr.Api.V3/Movies/MovieController.cs
plz12345, line 864: @sampulsar 
Need to set the default sort keys on this method as well, as it is used when a user is filtering the view by Tags.  That's why I suggested using a const similar to how we set `_AllowedSortKeys` in this class.

src/NzbDrone.Core.Test/Datastore/BasicRepositoryFixture.cs
```diff
@@ -324,5 +325,21 @@ public void get_paged_should_work_with_null_sort_key(int page, int count)
...
+        {
+            Subject.InsertMany(_basicList);
+            var pageSpec = new PagingSpec<ScheduledTask>() { Page = page, PageSize = 2, SortKey = "Interval", SortDirection = SortDirection.Descending, DefaultSortKeys = new List<string> { "LastExecution" } };
+            var data = Subject.GetPaged(pageSpec);
+
+            data.Page.Should().Be(page);
+            data.PageSize.Should().Be(2);
+            data.TotalRecords.Should().Be(_basicList.Count);
+            data.Records.Should().BeEquivalentTo(_basicList.OrderBy(x => x.LastExecution).OrderByDescending(x => x.Interval).Skip((page - 1) * 2).Take(2));
+            data.Records.Should().NotBeEquivalentTo(_basicList.OrderByDescending(x => x.LastExecution).OrderByDescending(x => x.Interval).Skip((page - 1) * 2).Take(2));
```
plz12345, line 342: @sampulsar 
This line should not be needed since you are already doing a positive assertion right above it.  It would be fragile, if the record LastExecution order changed unexpectedly.

plz12345 (changes requested):

src/NzbDrone.Core.Test/Datastore/BasicRepositoryFixture.cs
sampulsar: This is the point if the building of _basicList is altered so that Last Execution is not unique for some other reason later this test will then fail. It then stops a change after that breaking the code.

plz12345 (approved):

PR #252, opened 2026-03-29
Feat: Add posters view to PerformerDetails page

sk8ersteve:
#### Database Migration

NO

#### Description

Added PerformerDetailsPosters component which uses ScenePoster and MoviePoster to render poster views of all the performers works similar to SceneIndexPosters and MovieIndexPosters views. 

This is my second attempt pushing this change. I've reworked it to reuse the existing components as discussed and personally vetted and tested the code better. I did make sure to update the app state to include whatever view settings are set by the user. I'm not super familiar with typescript so a second set of eyes is definitely needed.

I made the decision to place movies below scenes rather than have them mixed in and force them to use the scene poster thumbnail aspect ratio, but I'm open to suggestions on better ways to deal with this.

<details>

20260329_15h21m01s_grim
20260329_15h21m16s_grim

</details>

#### Todos

- [x] Tested locally
- [ ] Consider making scenes/movies collapsible or toggleable. Right now movies are shown after scenes
- [ ] Consider making pageable like SceneIndex

- Fixes whisparr/whisparr#284

Note: not an exact fix. This change doesn't allow filtering by multiple performers (useful for performing bulk actions for multiple performers' scenes at once) but does allow viewing posters for one performer at a time (useful for browsing and deciding what to download).

### DIFF
--- a/frontend/src/App/State/AppState.ts
+++ b/frontend/src/App/State/AppState.ts
@@ -100,6 +100,12 @@ interface AppState {
   movieIndex: MovieIndexAppState;
   movieSearch: MovieSearchAppState;
   sceneIndex: MovieIndexAppState;
+  performerScenes: {
+    view: string;
+    posterOptions: {
+      size: string;
+    };
+  };
   performers: PerformersAppState;
   studios: StudiosAppState;
   movies: MoviesAppState;
--- a/frontend/src/Performer/Details/PerformerDetails.tsx
+++ b/frontend/src/Performer/Details/PerformerDetails.tsx
@@ -1,6 +1,6 @@
 import _ from 'lodash';
 import React, { useState } from 'react';
-import { useSelector } from 'react-redux';
+import { useDispatch, useSelector } from 'react-redux';
 import { useParams } from 'react-router-dom';
 import Flag from 'react-world-flags';
 import AppState from 'App/State/AppState';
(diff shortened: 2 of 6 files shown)


plz12345:
Still awaiting screenshots (use SFW mode)

sk8ersteve:
> Still awaiting screenshots (use SFW mode)

I'm confused. I've attached screenshots using the built in blur mode. Is that alright? The screenshots are visible once you tap the "details" element and it drops down. I believe that's the default for the template.

plz12345:
Ah, sorry.

So, does the View dropdown control this?  Is the default still the year-grouped table view?  Does the user's selection persist/survive a page reload?

sk8ersteve:
Yes the view lets you select between table and posters. When posters is selected, the options button replaces the expand all button. Currently the only option in options is poster size. And yes, this setting saves to the AppState.

plz12345:
In review, this appears to work as intended, but there is a performance issue.  Because I never designed the performer controller to handle cover images for movies/scenes, it is hotlinking those from upstream, rather than local cache.

In `MovieController`, this is done via `MapCoversToLocal` method call within`GetPagedMoviesStandard`.  This would need to be replicated for the `GetMoviesByPerformerForeignId` method in `PerformerController`.

plz12345:
You abandon this effort?

sk8ersteve:
Not completely. I'm extremely busy right now in personal life and this has turned into maybe a bigger endeavor than I originally anticipated but if I'm going to continue using whisparr, I'd like to eventually get this functionality. I'll probably be able to look at it again in like August.

plz12345:
Closing since this was abandoned.  User can re-create later if desired.

PR #300, opened 2026-06-08
refactor: modernize studio and performer lookups, nuke redux boilerplate

Mothball7205:
#### Database Migration

NO

#### Description

started this by trying to fix studio search spamming the endpoints. as soon as the input field wasn't empty, it would just loop spam the list endpoint

took a look at how other lookups were implemented. movie search was implemented way better because it uses react-query and determines DB existence directly on the backend search controller rather than dumping the whole collection client-side. performer search had the exact same over-fetching loop bug as studios

- updated backend search controller to map and return whether the studio/performer already exists
- migrated both useAddNewStudio and useAddNewPerformer hooks to react-query
- nuked all the legacy redux lookup/add actions, thunks, and reducers we don't need anymore (but kept the defaults state slices for settings persistence)

let me know if i missed something or if i broke any unexpected assumptions here

#### Todos

- [x] Tests
- [x] Translation Keys (./src/NzbDrone.Core/Localization/Core/en.json) (none modified/added)

Not reported

### DIFF
--- a/frontend/src/Performer/AddPerformer/AddNewPerformer.tsx
+++ b/frontend/src/Performer/AddPerformer/AddNewPerformer.tsx
@@ -1,5 +1,4 @@
 import React from 'react';
-import { Error } from 'App/State/AppSectionState';
 import Alert from 'Components/Alert';
 import TextInput from 'Components/Form/TextInput';
 import Icon from 'Components/Icon';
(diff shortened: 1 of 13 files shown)


plz12345:
Can you break this up into smaller PR's?  Like one for performer and one for Studio?  This is a lot of UX you're throwing a sledgehammer at, so splitting it reduces risk of breakage.

Mothball7205:
> Can you break this up into smaller PR's? Like one for performer and one for Studio? This is a lot of UX you're throwing a sledgehammer at, so splitting it reduces risk of breakage.

yeah you're right. i was thinking this might be too big. should have done that.
i will do that tomorrow or so

Mothball7205:
i will close this. one half of this PR is #301

PR #301, opened 2026-06-09
chore: migrate "add new studio" lookup and add flow to tanstack query

Mothball7205:
#### Database Migration

NO

#### Description

smaller part of https://github.com/Whisparr/Whisparr-Eros/pull/300

moves the studio lookup/add flow off redux and onto tanstack query, matching the pattern already used for movies. the isExisting flag now comes directly from the backend lookup endpoint instead of requiring a separate round-trip


plz12345:
Not sure if you're aware, this is both still in draft, and now has a merge conflict locking it.

Mothball7205:
> Not sure if you're aware, this is both still in draft, and now has a merge conflict locking it.

yes, i am aware

plz12345 (approved):
Looks to be copying the pattern we use on movies and scenes, so no major concerns.  Thanks for splitting into smaller PR chunks.  We can let this ride on eros-develop for a couple of weeks before looking to do a merge on the performer flavor of this.

PR #303, opened 2026-06-14
fix: stop add-studio/performer lookup from looping requests

Mothball7205:
#### Database Migration

NO

#### Description

fixes the infinite loop where typing in the "add new studio/performer" search field would spam /studio/list requests (the useEffect was re-triggering itself because it depended on the whole addMovie state object which it was also updating)
smallest possible fix


plz12345 (approved):
Solid gate to turn in-flight process into a no-op here, no concern.

PR #304, opened 2026-06-14
feat: add bulk delete endpoint for performers

Mothball7205:
#### Database Migration

NO

#### Description

adds a bulk delete endpoint for performers. currently we loop requests in the frontend when deleting more than one performer. will fix that later

let me know if I took everything into consideration for the delete logic


plz12345 (approved):
This looks solid, net-new endpoint, just moving the looping from client-side to server side.  Due to the cascade of movie/scene deletes, the performer deletion triggers, this is the right approach.

PR #338, opened 2026-07-20
Fix XBMC/Kodi NFO writing class name instead of studio title

Ai-Shihab:
#### Database Migration
NO

#### Description
The XBMC/Kodi metadata consumer was writing the fully-qualified class name of the `Studio` object (`NzbDrone.Core.MetadataSource.SkyHook.Resource.StudioResource`) into the `<studio>` element of generated `.nfo` files, instead of the actual studio title.

This happened because `movie.MovieMetadata.Value.Studio` is a `StudioResource` object, not a string. When passed directly into `new XElement("studio", ...)`, `XElement` calls `.ToString()` on it, and since `StudioResource` doesn't override `ToString()`, it falls back to `Object.ToString()`, which returns the type's fully qualified name.

This PR fixes it by reading `.Title` off the `Studio` object instead of passing the whole object, with a null-conditional to avoid an exception for movies with no studio set.

<details>
N/A — backend-only fix, no UI changes.
</details>

#### Todos
- [ ] Tests

- Fixes whisparr/whisparr-eros#1115

### DIFF
--- a/src/NzbDrone.Core/Extras/Metadata/Consumers/Xbmc/XbmcMetadata.cs
+++ b/src/NzbDrone.Core/Extras/Metadata/Consumers/Xbmc/XbmcMetadata.cs
@@ -285,7 +285,7 @@ namespace NzbDrone.Core.Extras.Metadata.Consumers.Xbmc
 
                 details.Add(new XElement("year", movie.Year));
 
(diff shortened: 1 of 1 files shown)


plz12345:
We duplicated efforts.  I saw you comment on the issue rather than PR, at first, so I raised a PR.  Closing this as it will be in the next dev merge via the other PR.

PR #351, opened 2026-07-20
chore: upgrade eslint to 10.7.0

plz12345:
#### Database Migration

NO

#### Description

Upgrades eslint from 9.39.4 to 10.7.0, doing the ecosystem migration by hand
rather than as a version bump — Dependabot's group PR (#341) couldn't, and
failed the frontend build.

The crash was `TypeError: scopeManager.addGlobals is not a function`. eslint 10
moved to eslint-scope 9 (whose `ScopeManager` gained `addGlobals`), but
`@babel/eslint-parser` 7.x hands back a scope manager built on eslint-scope 5,
so eslint 10 called a method that didn't exist and died on the first `.js` file.

`@babel/eslint-parser` 8 fixes it, but peers `@babel/core ^8` and needs
Node `^22.18` — that would drag a Babel 8 migration and a Node bump into an
eslint upgrade. It only existed to parse JSX/modern syntax in `.js` files, which
`@typescript-eslint/parser` (already used for `.ts`/`.tsx`, already working under
eslint 10) does too. This points the `.js` block at it and drops the dependency.
**Node stays on 20.**

eslint 10 also removed the `radix` rule's redundant-radix check. `as-needed` is
still in the schema but never read, so it now enforces plain `always` —
inverting it: `parseInt(x)` was fine and is now an error, `parseInt(x, 10)` was
an error and is now fine. This collapses the option to `radix: 'error'` and adds
the explicit radix at the 72 affected call sites (applied via the rule's own
suggestion fixer). All are CSS dimension strings, ids or version numbers, none
hex-prefixed, so behaviour is unchanged.

`eslint-plugin-react` and `eslint-plugin-import` still declare peer ranges
topping out at eslint 9 and warn on install, but both were verified working
against eslint 10.

Verified locally on Node 20: `yarn lint`, `yarn stylelint-linux` and
`yarn build --env production` all pass. Supersedes #341.

N/A — tooling only, no UI change.

#### Todos

- [x] Tests — N/A (lint/build tooling; the frontend CI job is the test)
- [x] Translation Keys (./src/NzbDrone.Core/Localization/Core/en.json) — none added

- N/A


plz12345:
@sampulsar or @Gykes 
This one supersedes the other bump PR's, as it was not dependabot directly.  Kind of a cascade of dep changes, but the majority are just to satisfy eslint 10 enforcement of `Number.parseInt()`.

Gykes (approved):

PR #374, opened 2026-07-25
Fix: Parse JAV catalog codes after release tags

h2nkm5vydg-oss:
#### Database Migration

NO

#### Description

Fixes JAV catalog-code parsing for release names where tracker/indexer prefix tokens appear before the catalog code.

Previously, the JAV parser only handled dash-separated catalog codes at the start of the title, so releases like `+++ [FHD] TEST-123 Some Title` and `[8KVR] MOCK-456 [VR] Some Title` did not populate `ParsedMovieInfo.Code`.

This change adds bounded detection for standalone dash-separated JAV catalog codes after prefix/tag tokens, while avoiding substring matches inside words such as `FOOABC-1234BAR`. It also aligns bracketed catalog-code parsing with existing tests for underscore and dot separators.

Not UI related.

#### Todos

- [x] Tests
- [x] Translation Keys (./src/NzbDrone.Core/Localization/Core/en.json) - Not required


plz12345 (changes requested):
Thanks for this — the goal is right and the test-vacuity fix is a genuinely good catch. But the regex changes cause three regressions that the existing suite doesn't catch. I reproduced all of them locally by running the parser against the base commit and this PR's head and diffing the output.

## Blocking

### 1. The new unanchored regex sits *above* the FC2 pattern and hijacks 4–5-digit FC2 codes

`ReportTitleRegex` is ordered and first-match-wins (`Parser.cs` ~L331). The new pattern is inserted at L141, directly before the `JAV-FC2` regex at L144, so it claims the title first and captures the wrong substring:

| title | base | this PR |
|---|---|---|
| `FC2-PPV-1234 Some Title` | `FC2-PPV-1234` | **`PPV-1234`** |
| `[FC2-PPV-4321] Title` | `FC2-PPV-4321` | **`PPV-4321`** |
| `FC2-PPV-1234567 Some Title` | `FC2-PPV-1234567` | `FC2-PPV-1234567` (unaffected — 7 digits can't satisfy `[0-9]{3,5}` plus the lookahead) |

Only 4–5-digit FC2 IDs break, which is exactly the older FC2 back catalogue. `Code` is used as a title-lookup fallback in `MovieService.cs` L689-692, so `PPV-1234` just fails to find the movie.

### 2. The trailing lookahead added to the anchored regex makes suffixed JAV codes unparseable entirely

Adding `(?=$|[\s._+\-\[\]\(\)\{\}])` to L140 means a code followed immediately by a letter no longer matches. `-A`/`-B` part splits and `-C` subtitled variants are common JAV naming, and the no-dash form is equally common:

| title | base | this PR |
|---|---|---|
| `SSIS-001A.mp4` | `SSIS-001` | **`ParseMovieTitle` returns `null`** |
| `SSIS-001C.mp4` | `SSIS-001` | **`null`** |
| `ABP-123A` | `ABP-123` | **`null`** |

That's not just a lost code — no regex matches at all, so the release/file becomes unparseable.

This lookahead also isn't needed for the stated goal. The `FOOABC-1234BAR` guard is satisfied by the *leading* boundary in the new L141 pattern alone (`^FOOAB` fails at `C`, and no interior position has a preceding separator). I verified `FOOABC-1234BAR` still yields `null` with L140 reverted to its original form.

### 3. The unanchored match invents codes on ordinary non-JAV scene names

Any `2–5 letters + dash + 3–5 digits` token anywhere in the title now becomes a catalog code:

| title | base | this PR |
|---|---|---|
| `Some.Studio.Part-123.1080p` | *(no code)* | **`Part-123`** |
| `[Vixen] Some Title - Part-234 - 1080p` | `null` | **`Part-234`** |
| `Brazzers - Some Title - Scene-123` | `null` | **`Scene-123`** |

This isn't cosmetic: a non-empty `Code` is one of the three conditions that lets a scene release through the parseability gate in `DownloadDecisionMaker.cs` L81 (and past the rejection at L116). Releases previously rejected as unparseable now enter the decision pipeline carrying a junk code that then drives `FindByTitle`.

### A shape that fits all four constraints

Requiring the code to follow a *closing bracket* (i.e. a tracker/tag block) rather than any separator satisfies both cases in the PR description without the collateral damage. Revert L140, move the new pattern below the FC2 regex, and use a lookbehind:

```csharp
// JAV
new Regex(@"^(?<code>[A-Z]{2,5}[- ][0-9]{3,5})", RegexOptions.IgnoreCase | RegexOptions.Compiled),

// JAV-FC2
new Regex(@"(?<code>FC2.*(?:PPV).*[0-9]{4,7})", RegexOptions.IgnoreCase | RegexOptions.Compiled),

// JAV code after leading tracker/tag blocks
new Regex(@"(?:^|(?<=[\]\)\}])[\s._+\-]*)(?<code>[A-Z]{2,5}-[0-9]{3,5})(?=$|[\s._+\-\[\]\(\)\{\}])", RegexOptions.IgnoreCase | RegexOptions.Compiled),
```

I ran this locally: it gives `TEST-123` and `MOCK-456` for the two cases you're fixing, keeps `FOOABC-1234BAR` at `null`, restores `FC2-PPV-1234` / `SSIS-001`, drops `Part-123` / `Part-234` / `Scene-123`, and the full `ParserTests` suite stays green. Offered as a starting point, not a mandate — there may be a shape you prefer.

## Should-fix

- **There is not a single FC2 test case anywhere in `ParserTests`** (`grep -rn FC2 src/NzbDrone.Core.Test/ParserTests/` returns nothing). That's why CI is green on #1. Whatever the final regex looks like, this PR should add FC2 cases to `should_correctly_parse_code`, since it edits the list position immediately above that pattern.
- No negative cases in the new coverage. `FOOABC-1234BAR` guards the *substring-inside-a-word* case, but nothing guards the *mid-title separated token* case. Please add e.g. `[TestCase("Some.Studio.Part-123.1080p", null)]` and `[TestCase("[Vixen] Some Title - Part-234 - 1080p", null)]`.
- No suffixed-code cases (`SSIS-001A`, `SSIS-001C`) to lock in #2.
- No linked issue in the body. Not mandatory, but if this tracks a report, please use `Fixes whisparr/whisparr#NNNN`.

## Worth keeping as-is

The test change at `ParseMovieTitleFixture.cs` L158-160 is the most valuable thing in this PR. `ParseMovieTitle(title)?.Code.Should().Be(result)` short-circuits the *entire* chain when the parse returns null, so every case that produced `null` was asserting nothing at all. Hoisting it into a local makes those assertions real — which is what exposed that `[MNO.678]` and `[JKL_345]` never actually worked, hence the `[- _.]` fix at L147. That bracket regex change is correct, and the header comment above it is now accurate rather than aspirational. Good catch.

## Checklist

- Base branch `eros-develop` — pass
- Head branch meaningfully named — pass
- Commit prefix (`Fix:`) — pass
- One fix per PR — pass
- DB-migration note (`NO`, no migration in diff) — pass
- Tests / Translation Keys checkboxes — pass
- `en.json` keys — n/a, no user-facing strings
- Linked issue — none
- Tests for changed backend code — partial; 3 cases added, but no FC2, negative, or suffix coverage, so the suite passes while three regressions ship
- Local verification — ran `ParserTests` on this PR's head: 1245 passed, 0 failed. The regressions above were found by differential probing against the base commit, not by the suite.
- CI: backend matrix (11), `frontend`, `unit_test` (3), `unit_test_postgres` (3), `integration_test` (3), `deploy/package` — all pass.
- CI: SonarQube "Build and Analyze" — fails, but **not this PR's fault**. `sonarqube.yml` runs under `pull_request_target` and dies at checkout with `Refusing to check out fork pull request code from a 'pull_request_target' workflow`; it never reaches analysis. That will fail on every fork PR and deserves a separate issue against the workflow.

h2nkm5vydg-oss:
Thanks for the detailed review!

### 1. FC2 precedence

> “hijacks 4–5-digit FC2 codes”

The regex order is now:

1. Original start-anchored JAV pattern
2. Existing FC2 pattern, unchanged
3. New post-tag JAV pattern

This prevents the generic matcher from claiming `PPV-1234`. Tests should now cover:

- `FC2-PPV-1234 Some Title`
- `[FC2-PPV-4321] Title`
- `FC2-PPV-1234567 Some Title`

All return the complete FC2 code.

### 2. Suffixed JAV codes

> “makes suffixed JAV codes unparseable entirely”

I restored the original start-anchored JAV regex without the trailing boundary, preserving the existing behavior:

- `SSIS-001A.mp4` → `SSIS-001`
- `SSIS-001C.mp4` → `SSIS-001`

The new post-tag matcher also supports a single-letter suffix while capturing only the base code:

- `[FHD] SSIS-001A.mp4` → `SSIS-001`

### 3. False positives in ordinary titles

> “invents codes on ordinary non-JAV scene names”

I removed the broad unanchored separator matcher.

The replacement is anchored to the beginning of the title and requires one or more complete leading `[...]`, `(...)`, or `{...}` tag blocks before the catalog code. This is deliberately stricter than matching after any closing bracket, since the latter could also match a code-like token after a later, mid-title bracket.

Negative coverage now includes:

- `Some.Studio.Part-123.1080p`
- `[Vixen] Some Title - Part-234 - 1080p`
- `Brazzers - Some Title - Scene-123`
- `Some Title [FHD] ABC-123`
- `FOOABC-1234BAR`
- `[FHD] ABC-1234BAR`

None populate `Code`.

### 4. Missing regression coverage

> “no FC2, negative, or suffix coverage”

All three categories are now covered in `should_correctly_parse_code`, including plain and bracketed FC2 IDs, suffixed codes, separated mid-title tokens, embedded-word tokens, and the stricter leading-tag boundary.

plz12345 (approved):
All three blocking regressions from my previous review are fixed. I re-verified each one by differential probing (base `a596f3dde6` vs head `71fc4be27a`) rather than by reading the diff.

## Blocking

None.

| Prior finding | Status | Evidence |
|---|---|---|
| **1.** New regex above `JAV-FC2`, hijacking 4–5-digit FC2 codes | **Fixed** | Pattern moved to `Parser.cs` L146, below FC2 at L143. `FC2-PPV-1234` → `FC2-PPV-1234`, `[FC2-PPV-4321]` → `FC2-PPV-4321`, `FC2-PPV-1234567` → full code. Identical to base. |
| **2.** Trailing lookahead on L140 broke suffixed codes | **Fixed** | L140 is byte-identical to base. `SSIS-001A.mp4` / `SSIS-001C.mp4` / `ABP-123A` → `SSIS-001` / `SSIS-001` / `ABP-123`, all matching base. `[FHD] SSIS-001A.mp4` → `SSIS-001` (new, via the `[A-Z]?` suffix lookahead — correct). |
| **3.** Unanchored match invented codes on ordinary scene names | **Fixed** | Now requires one or more leading `[...]`/`(...)`/`{...}` blocks before the code. `Some.Studio.Part-123.1080p`, `[Vixen] Some Title - Part-234 - 1080p`, `Brazzers - Some Title - Scene-123`, `[Blacked] ... - Part-123 - 2023 - 1080p`, `[EroticaX] Some Girl - Scene-01 - 1080p` all match base exactly — no code. |

Both target cases work: `+++ [FHD] TEST-123 Some Title` → `TEST-123` and `[8KVR] MOCK-456 [VR] Some Title` → `MOCK-456`, each `null` on base. `FOOABC-1234BAR` and `[FHD] ABC-1234BAR` still return `null`, so the embedded-word guard survives without the L140 lookahead.

Anchoring to *leading* tag blocks is a better call than the "after any closing bracket" lookbehind I suggested — it's what keeps `Some Title [FHD] ABC-123` from producing a mid-title false positive. Good.

## Should-fix

**1. Residual false positives: the code group is case-insensitive.** L146 inherits `RegexOptions.IgnoreCase`, so any mixed-case `Word-NNN` token immediately after a leading tag block becomes a catalog code:

| title | base | this PR |
|---|---|---|
| `[NF] Show-123 Title` | *unparseable* | `Show-123` |
| `[XXX] Cum-101 Whatever` | *unparseable* | `Cum-101` |
| `[Evil Angel] Rocco-101 Something` | *unparseable* | `Rocco-101` |

This is **not** a regression — those titles returned `null` on base, so nothing that previously worked has broken. That's why it's should-fix rather than blocking. But a non-empty `Code` is one of the conditions satisfying `ParsedMovieInfo.IsValid` (`ParsedMovieInfo.cs` L56), so these go from cleanly rejected to entering the decision pipeline with a junk code and no title.

Real JAV catalog codes are uppercase by convention, and making just that group case-sensitive is a one-token change:

```csharp
(?<code>(?-i:[A-Z]{2,5})-[0-9]{3,5})
```

I applied and ran it: all three false positives above drop back to `null`, every true positive is retained (`TEST-123`, `MOCK-456`, `SSIS-001`, `MILF-101`, `IPX-177`, `STARS-500`, `MIDE-800`), and the full `ParserTests` suite stays green. The only other case it drops is `[fhd] abc-123 title`, which isn't covered either way. `[Team-Skeet] MILF-101` still matches — genuinely ambiguous, fine to leave. Suggestion, not a mandate.

**2. Still no linked issue.** If this tracks a report, please use `Fixes whisparr/whisparr#NNNN`.

## Nits

- The new pattern captures no `title` group and sits above the `(480|…|1080)p` fallback, so `[FHD] ABC-123.Title.1080p` goes from `title='ABC-123 Title'` to code-only with an empty title. That's the same shape the existing anchored JAV regex at L140 already has for `ABC-123.Title.1080p`, so it's consistent rather than new — noting it only so it isn't a surprise later.
- `[1080p] BBC-100 Something` went from `title='['` (garbage) to `code=BBC-100`. Improvement.
- The `[- _.]` fix at L149 plus the assertion de-vacuuming at `ParseMovieTitleFixture.cs` L169 remain the most valuable part of this PR: `[SKMJ_649]` and `[SKMJ.649]` were returning `null` while their test cases asserted nothing at all. Both now genuinely pass.

## Checklist

- Base branch `eros-develop` — pass
- Head branch meaningfully named — pass
- Commit prefixes (`Fix:` ×2) — pass
- One fix per PR — pass
- DB-migration note (`NO`, no migration in diff) — pass
- Tests / Translation Keys checkboxes — pass
- `en.json` keys — n/a, no user-facing strings
- Linked issue — none (not mandatory)
- Test coverage for changed backend code — pass. 14 new `[TestCase]` rows for a 3-line parser change, covering FC2 (plain, bracketed, 7-digit), suffixed codes, embedded-word negatives, and mid-title negatives. Every gap I named last round is closed.
- Local verification — ran `ParserTests` on head: 1256 passed, 0 failed. A 55-title differential probe against the base commit confirms the three fixes and surfaced the residual above.
- CI — backend matrix (11), `frontend`, `unit_test` (3), `unit_test_postgres` (3), `integration_test` (3), `deploy/package` (11), CodeQL: all pass. SonarQube "Build and Analyze" is in progress on the current run (`Approve fork analysis` succeeded, so it should get past the checkout failure that killed it last round).
- CI — `triage` fails, **not this PR's fault**: `.github/labeler.yml` is still in labeler-v4 format while `0dfd4ac4b9` bumped `actions/labeler` to v7, which needs `changed-files:` / `any-glob-to-any-file:` nesting. Error: `found unexpected type for label 'Area: API'`. This fails on every PR against `eros-develop` and needs its own fix.

Thanks for the thorough turnaround on this — the added FC2, suffix, and negative coverage is exactly what was missing.

plz12345 (approved):

PR #416, opened 2026-08-07
Fix: Respect StashDB import list limits

h2nkm5vydg-oss:
#### Database Migration

NO

#### Description

StashDB import lists rounded configured limits down to whole 100-item pages. This caused limits below 100 to return no results and other non-page-aligned limits to return fewer results than requested.

I rounded the required page count up for all StashDB import list types, then trim the fetched results to the exact configured limit. Tests cover sub-page, exact-page, and cross-page limits.

<details>
<summary>Not applicable — backend-only change</summary>
</details>

#### Todos

- [x] Tests
- [x] Translation Keys (not required; no user-facing strings changed)

None.

### DIFF
--- a/src/NzbDrone.Core/ImportLists/StashDB/StashDBSettingsBase.cs
+++ b/src/NzbDrone.Core/ImportLists/StashDB/StashDBSettingsBase.cs
@@ -16,7 +16,8 @@ namespace NzbDrone.Core.ImportLists.StashDB
 
             RuleFor(c => c.Limit)
                 .GreaterThan(0)
-                .WithMessage("Must be integer greater than 0");
+                .WithMessage("Must be integer greater than 0")
+                .LessThanOrEqualTo(1000);
         }
     }
 
--- a/src/NzbDrone.Core/ImportLists/StashDB/StashDBSettingsBase.cs
+++ b/src/NzbDrone.Core/ImportLists/StashDB/StashDBSettingsBase.cs
@@ -16,8 +16,7 @@ namespace NzbDrone.Core.ImportLists.StashDB
 
             RuleFor(c => c.Limit)
                 .GreaterThan(0)
-                .WithMessage("Must be integer greater than 0")
-                .LessThanOrEqualTo(1000);
+                .WithMessage("Must be integer greater than 0");
         }
     }
 
@@ -37,7 +36,7 @@ namespace NzbDrone.Core.ImportLists.StashDB
         [FieldDefinition(0, Label = "Api Key", Privacy = PrivacyLevel.ApiKey, HelpText = "Your StashDB Api Key")]
         public string ApiKey { get; set; }
 
-        [FieldDefinition(1, Label = "Limit", HelpText = "Limit the number of movies to get")]
+        [FieldDefinition(1, Label = "Limit", HelpText = "Limit the number of movies to get, values above 1000 are capped at 1000")]
         public int Limit { get; set; }
 
         [FieldDefinition(2, Label = "Sort Date Descending", Type = FieldType.Select, SelectOptions = typeof(SceneSort), HelpText = "Descending sort by date style")]
--- a/src/NzbDrone.Core/ImportLists/StashDB/StashDBSettingsBase.cs
+++ b/src/NzbDrone.Core/ImportLists/StashDB/StashDBSettingsBase.cs
@@ -14,7 +14,6 @@ namespace NzbDrone.Core.ImportLists.StashDB
                 .NotEmpty()
                 .WithMessage("Api Key must not be empty");
 
-            // Limit not smaller than 1 and not larger than 100
             RuleFor(c => c.Limit)
                 .GreaterThan(0)
                 .WithMessage("Must be integer greater than 0");
--- a/src/NzbDrone.Core.Test/ImportListTests/StashDB/StashDBImportFixture.cs
+++ b/src/NzbDrone.Core.Test/ImportListTests/StashDB/StashDBImportFixture.cs
@@ -22,6 +22,7 @@ namespace NzbDrone.Core.Test.ImportListTests.StashDB
     public class StashDBImportFixture : CoreTest
     {
         private const int PageSize = 100;
+        private const int MaxResultsPerQuery = 1000;
 
         private List<int> _requestedPages;
         private int _reportedCount;
@@ -50,6 +51,8 @@ namespace NzbDrone.Core.Test.ImportListTests.StashDB
         [TestCase(typeof(StashDBPerformerImport), 200, 1000, new[] { 1, 1, 2 })]
         [TestCase(typeof(StashDBStudioImport), 200, 1000, new[] { 1, 1, 2 })]
         [TestCase(typeof(StashDBTagsImport), 200, 1000, new[] { 1, 1, 2 })]
+        [TestCase(typeof(StashDBFavoriteImport), 2000, 1500, new[] { 1, 1, 2, 3, 4, 5, 6, 7, 8, 9, 10 })]
+        [TestCase(typeof(StashDBTagsImport), 2000, 50000, new[] { 1, 1, 2, 3, 4, 5, 6, 7, 8, 9, 10 })]
         public void should_fetch_only_enough_pages_and_respect_limit(Type importType, int reportedCount, int limit, int[] expectedPages)
         {
             _reportedCount = reportedCount;
@@ -60,7 +63,7 @@ namespace NzbDrone.Core.Test.ImportListTests.StashDB
             import.Definition = definition;
 
             var result = import.Fetch();
-            var expectedCount = Math.Min(reportedCount, limit);
+            var expectedCount = Math.Min(Math.Min(reportedCount, limit), MaxResultsPerQuery);
             var expectedStashIds = Enumerable.Range(1, expectedCount).Select(index => $"scene-{index}");
 
             result.AnyFailure.Should().BeFalse();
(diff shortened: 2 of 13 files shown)


src/NzbDrone.Core/ImportLists/StashDB/StashDBSettingsBase.cs
```diff
@@ -14,7 +14,6 @@ public StashDBSettingsBaseValidator()
                 .NotEmpty()
                 .WithMessage("Api Key must not be empty");
 
-            // Limit not smaller than 1 and not larger than 100
```
h2nkm5vydg-oss, line 17: this comment seemed outdated since the limit can be over `100`, and validation only requires it to be greater than `0`

plz12345 (commented):
## Verdict: Approve-with-nits

The bug is real and the fix is correct on both halves. `maxPagesAllowed = _maxResultsPerQuery / _pageSize` floored to 0 for any limit under 100 — zero requests generated, zero results — and truncated non-aligned limits down to the nearest 100. Ceiling the page count and trimming the final list to `Settings.Limit` addresses both. Localized, no migration, no user-facing strings.

### Blocking

None.

### Should-fix

**1. The adjacent off-by-one is left in place — same class of bug, one word away from the fix.**

`StashDBFavoriteRequestGenerator.cs:63` (and the Performer/Studio/Tags twins):

```csharp
var pagesInResponse = (jsonResponse.Data.QueryScenes.Count / _pageSize) + 1;   // unchanged
var maxPagesAllowed = StashDBPagingHelper.GetPageCount(_maxResultsPerQuery, _pageSize);  // fixed
```

Failure scenario: StashDB reports `count: 200`, `PageSize` 100, limit 1000 → `pagesInResponse` = 3, `pages` = 3. Pages 1 and 2 are full so `IsFullPage` never breaks, and page 3 is requested and returns an empty scene list. That's a wasted authenticated round trip on every sync, for every exact multiple of 100. `StashDBPagingHelper.GetPageCount(Count, _pageSize)` is exactly the right call here, and the PR already imports the helper into all four files.

**2. Limits above 1000 are still silently capped, which the PR title claims to fix.**

`HttpImportListBase.cs:79` breaks out at `MaxNumResultsPerQuery = 1000` (a `const`, so not overridable). A user setting Limit = 1500 gets 1000 items; `Take(1500)` can't recover what was never fetched. Strictly better than before, so not blocking — but since `StashDBSettingsBase.cs:18` validates only `GreaterThan(0)`, nothing tells the user. A `LessThanOrEqualTo(1000)` rule would close the loop, and this PR is already editing that validator.

**3. Cross-page coverage only exists for one of the four list types.**

`StashDBImportFixture.cs:57-62` — Performer, Studio, and Tags get `limit: 1` only; the 100 and 101 rows are Favorite-only. Two more `[TestCase]` rows would cover the multi-page path per type.

### Nits

- **Test fidelity**: the mock returns the identical 100 scenes (`scene-1`…`scene-100`) for *every* request, so the `limit: 101` case asserts 101 items of which 100 are duplicate `StashId`s. It passes only because `CleanupListItems` doesn't dedupe. Varying the payload per page — and asserting the requested page numbers rather than just the call count — would catch a generator that requests page 1 twice, which the current `Verify(..., Times.Exactly(n))` cannot.
- **Reflection to set settings** (`StashDBImportFixture.cs:53-55`) is a pragmatic workaround for the CRTP generic on `StashDBSettingsBase<TSettings>`, but it silently no-ops into a `NullReferenceException` if a property is ever renamed. Acceptable; just noting it.
- **Latent trap for a future cleanup**: the Tags classes declare `namespace NzbDrone.Core.ImportLists.StashDB.Studio` (pre-existing, all four files under `Tags/`). The new fixture has no `...StashDB.Tags` using and compiles only because of that. Out of scope here, but whoever fixes the namespace will break this test.
- Removing the stale `// Limit not smaller than 1 and not larger than 100` comment is a good call — the validator never enforced 100, and limits above it worked.

### Local verification

Ran the touched area on the PR branch:

```
  --filter "FullyQualifiedName~ImportListTests.StashDB&TestCategory!=IntegrationTest&..."
Passed!  - Failed: 0, Passed: 10, Skipped: 0, Total: 10
```

I also confirmed the new fixture is a genuine regression test rather than a passing tautology, by reverting each half of the fix independently:

| Reverted | Result |
|---|---|
| `Fetch()` override removed | **5 failed** / 5 passed |
| `GetPageCount` → `_maxResultsPerQuery / _pageSize` | **5 failed** / 5 passed |

Both halves are load-bearing and covered.

### Checklist

| Gate | Status |
|---|---|
| Base branch `eros-develop` | ✅ pass |
| Head branch meaningfully named | ✅ pass (`codex/fix-stashdb-import-limit`) |
| Commit prefix | ✅ pass (`Fix:`) |
| One fix per PR | ✅ pass |
| DB migration note | ✅ pass (`NO`, no migration in diff) |
| Tests checkbox + tests present | ✅ pass (76-line fixture, verified failing without the fix) |
| Translation Keys / `en.json` | ✅ n/a — no user-facing strings added |
| ~80% new-code coverage | ✅ pass — both changed code paths exercised |
| CI | ✅ green — all 10 `backend` matrix legs, `frontend`, `unit_test` ×3, `unit_test_postgres` 16/17/18, `integration_test` ×3, CodeQL ×3 all pass. SonarQube "Build and Analyze" still in progress; `deploy / release` skipped (normal). |

One thing to watch: SonarQube hasn't finished, so the new-code quality gate is unconfirmed. Worth a glance before merging, though nothing in the diff looks likely to trip it.

Thanks for the clear PR description and for including tests — the sub-page / exact-page / cross-page split made the fix easy to verify.

h2nkm5vydg-oss:
Thanks for the review! i’ve updated the pagination logic, added 1,000 limit validation, and strengthened the tests

plz12345 (commented):
Thanks for the quick turnaround — all three items from the last pass are addressed, and two of them better than I suggested.

Rather than send you round again for a one-line change, I've pushed a commit to this branch (`cba0c43`). Details below so nothing is a surprise.

### What was still wrong, and it was my fault

I asked for `LessThanOrEqualTo(1000)` on the `Limit` rule without checking how stored settings are used. Provider definitions are revalidated on **every sync**, not just on save:

```
FetchAndParseImportListService.Fetch()      → _importListFactory.Enabled()
ImportListFactory.Enabled()                 → GetAvailableProviders()
ProviderFactory.GetAvailableProviders():91  → Active()
ProviderFactory.Active():184                → All().Where(c => c.Settings.Validate().IsValid)
```

So a user with a saved StashDB list at `Limit = 5000` — which today syncs fine and returns 1000 items — would have had that list silently filtered out of `Active()` after this merged. No log line, no health check, no error in the UI; the list still shows as Enabled while quietly importing nothing. That's a worse failure mode than the over-limit behaviour the rule was meant to surface, and the rule was my suggestion, not a flaw in your work.

### What the commit does

- Removes `.LessThanOrEqualTo(1000)`, so existing definitions keep validating.
- Adds `StashDBImportBase.EffectiveLimit => Math.Min(Settings.Limit, MaxNumResultsPerQuery)`, used in `Fetch()` and passed to all four request generators.
- Updates the `Limit` field help text to say values above 1000 are capped.
- Adjusts `limit_should_validate_within_supported_range` (`1001` and `50000` are now valid, with a comment on why rejecting them is wrong) and adds two `should_fetch_only_enough_pages_and_respect_limit` rows covering limits above the ceiling.

Using the inherited `MaxNumResultsPerQuery` const instead of a fresh `1000` literal also resolves the magic-number nit from my last comment — the two can't drift apart now.

### One caveat on that commit

The clamp is **defensive, not observable** at the current `PageSize`. I checked by reverting `EffectiveLimit` to `Settings.Limit`, and all 21 tests still passed: `PageSize` is 100 and `MaxNumResultsPerQuery` is 1000, so the fetch loop's own `pagedMovies.Count >= MaxNumResultsPerQuery` break already lands exactly on a page boundary. The two new test rows pin the user-visible contract — limit 1500 or 50000 yields 1000 items and 10 page requests — but they don't independently discriminate the clamp itself.

I kept it anyway because it makes `Fetch()` correct if `PageSize` ever stops dividing 1000 evenly (at `PageSize` 300 you'd get 1200 items back without it), and it stops the generator building ~490 unused `ImportListRequest` objects per sync at `Limit = 50000`, each doing a full URL build. If you'd rather keep the diff minimal, dropping `EffectiveLimit` and keeping only the validator and help-text changes is perfectly defensible — your call, and I won't push back on it.

### Verification

```
  --filter "FullyQualifiedName~ImportListTests&TestCategory!=IntegrationTest&..."
Passed!  - Failed: 0, Passed: 36, Skipped: 0, Total: 36

dotnet build src/NzbDrone.Core/Whisparr.Core.csproj  →  0 Warning(s), 0 Error(s)
```

### Credit where it's due

Two things in your last commit were better than what I proposed:

`GetPageCount(Math.Min(Count, _maxResultsPerQuery), _pageSize)` takes the min *before* the ceiling divide, which collapses three lines into one and is subtler than swapping the helper in and keeping the outer `Math.Min` — I'd have left the clumsier version in place.

And the fixture rework is a genuine upgrade: varying the payload per page, asserting the exact requested page *sequence* rather than a call count, and `OnlyHaveUniqueItems()` all catch classes of bug the original couldn't. The `count: 200 / limit: 1000` rows pin the wasted-request fix precisely.

Nothing else outstanding from me — happy with this once CI is green on the new commit.

plz12345 (approved):

PR #429, opened 2026-08-09
Fix: Handle Skyhook Timeout and continue refreshing the remaining performers/studios

sampulsar:
#### Database Migration

NO

#### Description

Skip a studio/performer if skyhook timeout happens, and only get the works if it is going to be used.

Do not return empty StudioWorksResource data. Do not delete anything. Do not change the timeout yet.

This makes a timeout on:

https://api.whisparr.com/v4/site/0f1d8e2c-627e-4d05-a52e-c821b26cab0c/works

mean:

New Sensations → metadata timeout → log → skip
Studio 2        → continue
Studio 3        → continue
Studio 4        → continue
...

<details>

</details>

#### Todos

- [x] Tests

### DIFF
--- a/src/NzbDrone.Core.Test/PerformerTests/RefreshPerformerServiceFixture.cs
+++ b/src/NzbDrone.Core.Test/PerformerTests/RefreshPerformerServiceFixture.cs
@@ -9,6 +9,7 @@ using NUnit.Framework;
 
 using NzbDrone.Common.Http;
 
+using NzbDrone.Core.Messaging.Commands;
 using NzbDrone.Core.MetadataSource;
 using NzbDrone.Core.Movies;
 using NzbDrone.Core.Movies.Performers;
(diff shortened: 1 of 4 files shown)


plz12345 (changes requested):
Thanks for picking this up — the direction is right and the change is safe. But as written it only covers one of the ways the Skyhook call actually fails, so the title's promise ("continue refreshing the remaining performers/studios") doesn't hold for the more likely real-world variants. Two of the three code findings are one-line fixes.

## Blocking

### 1. A 5xx from the works endpoint still aborts the entire refresh

`RefreshStudioService.cs:121`, `RefreshPerformerService.cs:170`

`GetStudioWorks` sets `SuppressHttpError = true` and then throws itself (`SkyHookProxy.cs:544-551`): `MovieNotFoundException` on 404, **`HttpException` on everything else**. A `WebException` only arises when *our* client gives up first — `ManagedHttpDispatcher.cs:144` wraps the cancelled request as `WebException(WebExceptionStatus.Timeout)`.

Failure scenario: the API is slow and Cloudflare returns **504** (or the origin returns 502/503) before our client timeout fires. `HasHttpError` is true, the status isn't 404, so `HttpException` is thrown. The new `catch (WebException)` doesn't match. It propagates out of `SyncStudioItems`, past `catch (MovieNotFoundException)` at `RefreshStudioService.cs:341`, and kills `Execute` — New Sensations fails, studios 2..N never run. Exactly the symptom this PR is fixing, just via the neighbouring code path.

Same for `MovieNotFoundException`: a 404 on `/works` (as opposed to `/site/{id}`) escapes `SyncStudioItems` in the **ID-list** path, where `RefreshStudioService.cs:312` calls it outside any try/catch, so the remaining requested IDs get dropped.

### 2. A timeout on the *info* call still aborts the bulk run

`RefreshStudioService.cs:330-344`, `RefreshPerformerService.cs:372-386`

The scheduled (no-IDs) path guards `RefreshStudioInfo` + `SyncStudioItems` with `catch (MovieNotFoundException)` only. `GetStudioInfo` / `GetPerformerInfo` hit `/site/{id}` and `/performer/{id}` with identical error handling, so if the timeout lands on the *info* fetch instead of the *works* fetch, the whole command still dies at studio 1. The API is timing out on one host — there's no reason the info call would be spared.

**Fix for both, and it's the pattern already used in this repo** — mirror `RefreshMovieService.cs:377-387` and add a general handler to the loop bodies in `Execute`:

```csharp
catch (MovieNotFoundException)
{
    _logger.Error("Studio '{0}' (StashDb {1}) was not found, ...", studio.Title, studio.ForeignId);
}
catch (Exception e)
{
    _logger.Error(e, "Couldn't refresh info for {0}", studio.Title);
}
```

That single addition subsumes the `try/catch` inside `SyncStudioItems` entirely — timeout, 5xx and malformed JSON all skip one studio and continue — and it fixes the ID-list path too. If you'd rather keep the targeted catch inside `SyncStudioItems`, at minimum broaden it to `catch (Exception ex)` (or `catch (Exception ex) when (ex is WebException or HttpException)`), and still wrap the bulk loop.

## Should-fix

### 3. The `Monitored || MoviesMonitored` guard is dead code

`RefreshStudioService.cs:116`, `RefreshPerformerService.cs:165`

`SyncStudioItems` already returns at line 103 when `!Monitored && !MoviesMonitored`, so by line 116 the condition is unconditionally true. "Only get the works if it is going to be used" is a no-op here — no call is avoided that wasn't already avoided. It also forces the unused `(new List<string>(), new List<string>(), new List<int>())` initializer at line 112. Drop the `if` and declare without initialising:

```csharp
(List<string> StashdbIds, List<string> TpdbIds, List<int> TmdbIds) studioWork;

try
{
    studioWork = _movieInfo.GetStudioWorks(studio.ForeignId);
}
catch (...)
```

Definite assignment is satisfied because the catch returns.

### 4. Test for the new behaviour only

The Todos box is honestly left unchecked, and there's currently no fixture at all for either service — so to be clear, **I'm not asking for a test harness around the Skyhook proxy or general coverage for these services**. Just the new code path, bare bones:

- a `RefreshStudioServiceFixture : CoreTest<RefreshStudioService>`
- `Mocker.GetMock<IProvideMovieInfo>().Setup(x => x.GetStudioWorks(failingId)).Throws(new WebException("Http request timed out", WebExceptionStatus.Timeout))`
- assert `AddMovies` is still called for the studios after the failing one

That's the whole thing, and it's the only way to actually verify "the loop continues" — it can't be eyeballed. If you parameterise the exception type it pins finding #1 at the same time.

### 5. Placeholder linked issue

The body still reads `Fixes whisparr/whisparr#XXXX`. Fill in the real issue number or drop the line so it doesn't render as a broken link.

## Nits

- Continuation indent on the `_logger.Warn(` calls — `RefreshStudioService.cs:123-127` and `RefreshPerformerService.cs:172-176`. The arguments sit at the same indent level as the call itself. Either put it on one line (it fits) or indent the arguments one level.
- Nice cleanup removing the stale `// var studioWork = ...` comment.

## Checklist

| Gate | Result |
|---|---|
| Commit prefix | pass — `Fix:` |
| One concern per PR | pass |
| DB migration note | pass — `NO`, no migration in the diff |
| Linked issue | `#XXXX` placeholder |
| Tests present | none — see #4, new path only |
| `en.json` keys | n/a — log messages only |
| CI | all green: backend x10 RIDs, `frontend`, `unit_test` x3, `unit_test_postgres` 16/17/18, `integration_test` x3, CodeQL, all packages. SonarQube "Approve fork analysis" pending maintainer approval, so the new-code coverage gate never ran. |

plz12345 (approved):

PR #433, opened 2026-08-12
Fix: Year filtering on scene page referencing Movie column that does not exist

asusguy94:
#### Database Migration

NO

#### Description

Fix a network error that happens when filtering on _year_. Seems like we filter on `Movie.Year`, but _Movie_ does not have such a column, so we get an exception about this. This PR changes this to instead use the year property on _MovieMetadata_ (as we already do on other filters like "genres" and "runtime")

<details>
image

image

```json
{
    "message": "SQL logic error\nno such column: Movies.Year",
    "description": "code = Error (1), message = System.Data.SQLite.SQLiteException (0x87AF001F): SQL logic error\nno such column: Movies.Year\n ...
 }
```
</details>

#### Todos

- [ ] Tests

### DIFF
--- a/src/Whisparr.Api.V3/Movies/Helpers/Paging.cs
+++ b/src/Whisparr.Api.V3/Movies/Helpers/Paging.cs
@@ -78,7 +78,7 @@ namespace Whisparr.Api.V3.Movies.Helpers
                         break;
 
                     case "year":
-                        NumericFilterInt.Apply(pageSpec, jsonElement, op, p => p.Year);
+                        NumericFilterInt.Apply(pageSpec, jsonElement, op, p => p.MovieMetadata.Value.Year);
                         break;
                 }
             }


plz12345 (approved):
Nice catch — this is correct, and it matches the pattern already used by the sibling filters in the same file. I verified the fix end-to-end locally rather than just reading the diff.

## Root cause confirmed

`TableMapping.cs` explicitly excludes the property:

```csharp
Mapper.Entity<Movie>("Movies").RegisterModel()
      .Ignore(s => s.RootFolderPath)
      .Ignore(s => s.Title)
      .Ignore(s => s.Year)      // <-- not a column
      ...
      .HasOne(a => a.MovieMetadata, a => a.MovieMetadataId);
```

`Movie.Year` is a C# compatibility property proxying to `MovieMetadata.Value.Year`. It reads fine in memory, but `WhereBuilderSqlite` emits `Movies.Year` and SQLite rejects it — exactly the error in your report.

## Reproduced and fixed, against a real DB

I wrote a throwaway `DbTest` fixture, ran it, then deleted it:

| Expression | Result |
|---|---|
| `m => m.Year == 2020` (pre-fix) | throws `SQLiteException: SQL logic error / no such column: Movies.Year` |
| `m => m.MovieMetadata.Value.Year == 2020` | passes — 1 record, `TotalRecords` = 1 |
| `m => years.Contains(m.MovieMetadata.Value.Year)` (the `equal` op) | passes — correct 2 of 3 |
| `m => m.MovieMetadata.Value.Year > 2019` (`greaterthan`) | passes — correct 2 of 3 |

I covered `Contains` and `greaterthan` deliberately, since `NumericFilterInt` builds a different expression shape per operator and `TotalRecords` comes from a separate `SelectCount()` query — both translate correctly.

No regressions: `MovieRepositoryTests` + `PagedFilters` + `MovieServiceTests` — 93 passed, 0 failed.

## Correct for both code paths

`PagedBuilder()` INNER JOINs `MovieMetadata`, so `MovieMetadata.Year` resolves in the SQL path. In the in-memory tag-filter path (`GetPagedMoviesWithTags`) the expression is compiled and run over `GetAllMovies()`, where behavior is unchanged — `m.Year` was already dereferencing `MovieMetadata.Value.Year` internally. No regression either way.

## Adjacent code audited — no sibling bugs

I checked every other filter key in `Paging.cs` against the `Ignore()` list. `Added`, `Monitored`, `Path`, `QualityProfileId` are real `Movies` columns; `Genres`, `ItemType`, `Status`, `Runtime`, `Title` already go through `MovieMetadata`. `ReleaseDateUtc` is a genuine mapped column on `MovieMetadata` (only `Credits` and `TagIds` are ignored there). `Year` was the last one referencing an ignored property.

This also brings `Paging.cs` in line with `Sorting.cs`, which already maps `"year"` → `"movieMetadata.Year"`. Sorting by year was never broken; only filtering was — consistent with the symptom you saw.

## One suggestion (non-blocking)

Would you add a regression test? The scaffolding is already there — `MovieRepositoryFixture` is a `DbTest<MovieRepository, Movie>` and already has `GivenPagedMovie(profileId, movieFileId, year = 2020)` with a `year` parameter plus a `PagingSpec()` helper, so it's about six lines:

```csharp
[Test]
public void should_filter_paged_movies_by_metadata_year()
{
    var profile = GivenProfile();
    var movieFile = GivenMovieFile(Quality.HDTV720p);
    GivenPagedMovie(profile.Id, movieFile.Id, 2020);
    GivenPagedMovie(profile.Id, movieFile.Id, 2021);

    var spec = PagingSpec();
    spec.FilterExpressions.Add(m => m.MovieMetadata.Value.Year == 2020);

    Subject.GetPaged(spec).Records.Single().MovieMetadata.Value.Year.Should().Be(2020);
}
```

Worth stressing why it has to go through `GetPaged`: the existing `Datastore/PagedFilters/` fixtures compile the expression and evaluate it in-memory against a POCO, where `m.Year` works fine. An in-memory filter test would be false comfort here — only the SQL path fails.

## Checklist

- Base branch `eros-develop` — pass
- Commit prefix `Fix:` — pass
- DB migration note (`NO`) — pass, no migration in diff
- Translation keys — N/A, no new user-facing strings
- Tests — none added (see suggestion above)
- CI — all green: 10/10 backend matrix, `frontend`, `unit_test` (3 OS), `unit_test_postgres` (16/17/18), `integration_test` (3 OS), SonarQube "Build and Analyze", CodeQL

Approving — the fix is unambiguously right. I'd merge with the test added, but I won't hold it up over that.

PR #448, opened 2026-08-16
Fix: Reuse compiled regex and parsed custom formats

ghost:
#### Database Migration

NO

#### Description

Two independent allocations removed from per-item paths. Neither changes behaviour.

`IdentificationService.Identify` constructed a `new Regex(..., RegexOptions.Compiled, ...)` on every call, and it runs once per file during scene disk scans. `RegexOptions.Compiled` emits IL at construction, so this was the expensive variant. Moved to a `static readonly` field — pattern, options and timeout are unchanged.

`UpgradeDiskSpecification.IsSatisfiedBy` computed `_formatService.ParseCustomFormat(file)` into a local for logging, then called it a second time as an argument to `CutoffNotMet`, re-running every custom-format specification against the file. It now passes the local it already had.

N/A — no UI changes.

#### Todos

- [x] Tests — no new tests; neither change alters behaviour (identical regex construction args, and `ParseCustomFormat` is pure over the same unchanged `file`, computed before any branch). Existing suite green.
- [x] Translation Keys — N/A, no user-facing strings.

- N/A

### DIFF
--- a/src/NzbDrone.Core/DecisionEngine/Specifications/UpgradeDiskSpecification.cs
+++ b/src/NzbDrone.Core/DecisionEngine/Specifications/UpgradeDiskSpecification.cs
@@ -44,7 +44,7 @@ namespace NzbDrone.Core.DecisionEngine.Specifications
 
             if (!_upgradableSpecification.CutoffNotMet(qualityProfile,
                     file.Quality,
-                    _formatService.ParseCustomFormat(file),
+                    customFormats,
                     subject.ParsedMovieInfo.Quality))
             {
                 _logger.Debug("Cutoff already met, rejecting.");
--- a/src/NzbDrone.Core/MediaFiles/MovieImport/Identification/IdentificationService.cs
+++ b/src/NzbDrone.Core/MediaFiles/MovieImport/Identification/IdentificationService.cs
@@ -26,6 +26,11 @@ namespace NzbDrone.Core.MediaFiles.MovieImport
 
     public class SceneIdentificationService : ISceneIdentificationService
     {
+        private static readonly Regex FolderDateRegex = new Regex(
+            @"(?<airyear>\d{2}|\d{4})[-_. ]+(?<airmonth>[0-1][0-9])[-_. ]+(?<airday>[0-3][0-9])",
+            RegexOptions.IgnoreCase | RegexOptions.Compiled,
+            RegexDefaults.Timeout);
+
         private readonly ISearchForNewMovie _searchProxy;
         private readonly IAddMovieService _addMovieService;
         private readonly IAggregationService _aggregationService;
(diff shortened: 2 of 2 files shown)


plz12345 (changes requested):
The code here is sound — I verified both hunks rather than taking the description on trust, and every technical claim in it holds up. My request is about packaging, not correctness.

## Requested change: split this into two PRs

This bundles two unrelated changes in two unrelated subsystems — a regex hoist in `MediaFiles/MovieImport/Identification`, and a duplicate-call removal in `DecisionEngine/Specifications`. They share only the theme "remove redundant per-item work."

I know #447 was split four ways to address exactly this, and I appreciate that. But the split was made by *theme* (perf) rather than by *subsystem*, which is why this one still straddles two areas. One change per PR, please — they'll both go in quickly on their own, and it keeps the history bisectable if either ever needs reverting independently.

Nothing below blocks either half once separated.

## What I verified (both hunks are fine)

**The regex hoist is correct, and the reasoning behind it is sharper than it looks.**

`RegexOptions.Compiled` triggers reflection-emit of IL *at construction time*, so building it per call was strictly worse than not passing `Compiled` at all — you paid the codegen cost every call and discarded the result. Calling that out as "the expensive variant" is the right diagnosis, not just pattern-matched "hoisting is good."

I confirmed the hot path: `Identify` is reached from `ImportDecisionMaker.GetRemoteMovie(string file)`, once per file during scene disk scans. Pattern, options and timeout are byte-identical to the original, and the declaration matches house convention — PascalCase `private static readonly Regex` with `RegexDefaults.Timeout`, consistent with `TorrentRssParser`, `FileNameBuilder`, `SearchMovieComparer` and the rest of the #441 timeout work.

**The custom-format reuse is behaviour-preserving.**

`UpgradeDiskSpecification.cs:41` already computed `_formatService.ParseCustomFormat(file)` into `customFormats`, and line 47 called it again with the identical argument — only `file.Movie = subject.Movie` happens in between, and that's before both.

I specifically checked the one thing that could have made this a real bug: aliasing. Passing the same `List<CustomFormat>` instance to both `CutoffNotMet` and the later `IsUpgradable`/`CalculateCustomFormatScore` would break if any consumer mutated it. None do — `CustomFormatCutoffNotMet` only reads, and `CalculateCustomFormatScore` is `FormatItems.Where(x => formats.Contains(x.Format)).Sum(x => x.Score)`. The saving is also real rather than cosmetic, since `ParseCustomFormat` runs every custom-format specification against the file.

## Nits (non-blocking)

- **`Fix:` prefix.** Same note as on #450 — the description is explicit that neither change alters behaviour, so `Chore:` or `refactor:` describes it better.
- **`RegexOptions.IgnoreCase` is a no-op on that pattern** — it's `\d`, `[-_. ]` and digit classes, no letters. Pre-existing, and correctly left alone by a PR that promises identical options. Just noting it for whoever touches it next.
- **`SceneIdentificationService` has no test fixture anywhere in the repo.** Your "no new tests" waiver is justified for a pure declaration hoist, so this isn't your debt — but it does mean nothing guards that regex. Worth someone filing separately.

## Verification

```
  --filter "FullyQualifiedName~DecisionEngineTests&TestCategory!=IntegrationTest&..."

Passed! - Failed: 0, Passed: 255, Skipped: 1, Total: 256, Duration: 589 ms
```

255 decision-engine tests green including `UpgradeDiskSpecificationFixture`. The identification hunk has no fixture to run.

CI note: `triage` passed, "Approve fork analysis" is pending the standard fork-PR approval gate — not a fault in this PR, but `build_v3` and Sonar haven't reported yet.

PR #449, opened 2026-08-16
Fix: Use hash lookups for import list sync comparisons

ghost:
#### Database Migration

  NO

  #### Description

  `ImportListSyncService` used linear scans to compare list items against the library, making a sync O(list items × library). `ProcessMovieReport` probed `dbMovies` with `List.Contains` per report and scanned `moviesToAdd` with `.All(...)` per accepted item; `CleanLibrary` ran `listMovies.Any(...)` per library movie. All three are now hash lookups.

  Notes:

  - The exclusion check moved from `SingleOrDefault` to `HashSet.Contains`. `SingleOrDefault` throws on duplicates, but `ImportExclusions.ForeignId` is `UNIQUE` (`001_initial_setup.cs:219`), so that path was unreachable.
  - `pendingForeignIds.Add(...)` replaces `moviesToAdd.All(...)`, with the same result: the body adds exactly one movie per accepted id, and reports are already deduped by `DistinctBy` upstream.

  N/A

  #### Todos

  - [ ] Tests: none yet, as there is no existing `ImportListSyncService` fixture. Can add one before merge.
  - [x] Translation Keys: N/A

  - N/A

### DIFF
--- a/src/NzbDrone.Core/ImportLists/ImportListSyncService.cs
+++ b/src/NzbDrone.Core/ImportLists/ImportListSyncService.cs
@@ -78,7 +78,7 @@ namespace NzbDrone.Core.ImportLists
             ProcessListItems(listItemsResult);
         }
 
-        private void ProcessMovieReport(ImportListDefinition importList, ImportListMovie report, List<ImportListExclusion> listExclusions, List<string> dbMovies, List<Movie> moviesToAdd)
+        private void ProcessMovieReport(ImportListDefinition importList, ImportListMovie report, HashSet<string> excludedForeignIds, HashSet<string> dbMovies, List<Movie> moviesToAdd, HashSet<string> pendingForeignIds)
         {
             if (report.ForeignId.IsNullOrWhiteSpace() || !importList.EnableAuto)
             {
@@ -93,16 +93,14 @@ namespace NzbDrone.Core.ImportLists
             }
 
             // Check to see if movie excluded
-            var excludedMovie = listExclusions.SingleOrDefault(s => s.ForeignId == report.ForeignId);
-
-            if (excludedMovie != null)
+            if (excludedForeignIds.Contains(report.ForeignId))
             {
                 _logger.Debug("{0} [{1}] Rejected due to list exclusion", report.ForeignId, report.Title);
                 return;
             }
 
             // Append Artist if not already in DB or already on add list
-            if (moviesToAdd.All(s => s.ForeignId != report.ForeignId))
+            if (pendingForeignIds.Add(report.ForeignId))
             {
                 var monitorType = importList.Monitor;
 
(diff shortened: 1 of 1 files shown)


plz12345 (commented):
This is the best-packaged of the four — one file, one idea, no scope creep. The code is correct on all three conversions; I verified each rather than taking the description on trust. My notes are about the description and one coverage gap, not the implementation.

## The code is right

**`SingleOrDefault` → `HashSet.Contains`.** The justification checks out precisely. `ImportExclusions.ForeignId` is `.NotNullable().Unique()` at exactly `001_initial_setup.cs:219` — the cited line is exact. I also checked the only two later migrations touching that table (`005_import_exclusion_type`, `014_import_exclusion_reason`); both only *add* columns, so the constraint stands. The unique constraint is on `ForeignId` alone rather than `(ForeignId, Type)`, so there's no type-scoped duplicate hiding either. `SingleOrDefault`'s throw path really was unreachable.

**`moviesToAdd.All(...)` → `pendingForeignIds.Add(...)`.** This is the one that could have been a real bug, and it isn't. The `if` body unconditionally adds exactly one `Movie` with `ForeignId = report.ForeignId` — no early returns, no nested conditionals — so `pendingForeignIds` stays exactly in sync with the ForeignId set of `moviesToAdd`. Lifetimes match too: both are created fresh in `ProcessListItems` and shared across every list in the grouped loop, so cross-list dedup is preserved.

**`CleanLibrary`'s `Any(...)` → hoisted `HashSet`.** Same ordinal comparison, null-safe in both directions, projection correctly hoisted outside the `moviesInLibrary` loop.

## The fixture you're looking for does exist

> Tests: none yet, as there is no existing `ImportListSyncService` fixture.

`src/NzbDrone.Core.Test/ImportListTests/ImportListSyncServiceFixture.cs` — 14 tests, and it covers most of what you changed:

- `should_not_add_movie_from_on_exclusion_list` → the `SingleOrDefault` path
- `should_not_add_movie_that_exists_in_library` → the `dbMovies` path
- five `..._clean_library_...` tests → the `CleanLibrary` path

I suspect the miss is that the namespace is `NzbDrone.Core.Test.ImportList` while the directory is `ImportListTests/`, so grepping either one alone misses it. Worth knowing for future PRs — and worth double-checking claims about what *doesn't* exist in the repo before they go in a description, since here it's the stated reason for leaving the Tests box unchecked.

All 60 import-list tests pass on your branch, so this is already covered. Please just correct the note and tick the box.

## One real gap that is worth a test

Nothing currently distinguishes `pendingForeignIds.Add(...)` from the old `moviesToAdd.All(...)`.

`should_not_add_duplicate_movies_from_separate_lists` looks like it covers this, but it doesn't — it sets `_importListFetch.Movies[0].ForeignId = "ForeignId4"`, which collides with `Movies[3]` *within list 1*, and the `DistinctBy` at the top of `ProcessListItems` removes it before `ProcessMovieReport` ever sees it. The expected count of 7 = (5−1) + 3 confirms the dedup happened upstream. So that guard is dead code under current inputs, in both the old and new form.

A case with the same ForeignId in two *different* `ListId` groups would actually exercise it, and would pin the behaviour you're relying on in the description. Since you noted the `DistinctBy` interaction yourself, you clearly already understand this one — it just needs a test.

## Nits

- **`Fix:` prefix** on a change the description says doesn't alter behaviour. Same note as #448/#450 — `Chore:` or `refactor:` fits better. Not a gate failure, the repo takes both styles.
- The description is indented two spaces from `#### Database Migration` onward, so GitHub renders the whole thing as a code block and the template checkboxes come out unreadable.

## Verification

```
  --filter "FullyQualifiedName~ImportList&TestCategory!=IntegrationTest&..."

Passed! - Failed: 0, Passed: 60, Skipped: 0, Total: 60, Duration: 821 ms
```

CI is running on this one now — `prepare`, `triage` and CodeQL actions/js-ts green, backend matrix and `frontend` still pending, `unit_test` not yet spawned.

PR #450, opened 2026-08-16
Fix: Use hash lookups and batch movie fetch in refresh services

ghost:
#### Database Migration

NO

#### Description

Six blocks across `RefreshStudioService` and `RefreshPerformerService` matched every work against whole-library ID lists with `List.Contains`, inside deferred LINQ enumerated up to three times. Now `HashSet`, materialised once.

The per-entity re-read of `AllMovieStashIds()` is kept on purpose, because `AddMovies` runs inside that loop and each entity needs to see the previous one's additions. Only the lookup structure changed.

Both `Execute` methods also fetched already-loaded items one at a time with `GetMovie(item.Id)`. Now a single `GetMovies(...)`.

Notes:

- `.Distinct()` is required, not cosmetic: `BasicRepository.Get` throws when `result.Count != ids.Count()`. It also dedupes rescans, since `GetByPerformerForeignId` joins credits, so a performer with two credits on one movie used to get rescanned twice.
- If a movie is deleted mid-refresh, the exception changes from `ModelNotFoundException` (after earlier rescans) to `ApplicationException` (before any). Both propagate out of `Execute`.
- Rescan order may differ: repository order rather than `items` order.
- The TMDB exclusion projection no longer closes over a shared `tmbdId`. The original is correct, since LINQ streams element by element, so this is readability rather than a fix.
- Dropped a now-unused `using DryIoc.ImTools;`, as unnecessary usings are build errors here.

N/A

#### Todos

- [x] Tests: `RefreshPerformerServiceRescanFixture` covers `GetMovies` called once with distinct ids, `GetMovie` never called, and no fetch when `items` is empty.
- [x] Translation Keys: N/A

- N/A

### DIFF
--- a/src/NzbDrone.Core.Test/PerformerTests/RefreshPerformerServiceRescanFixture.cs
+++ b/src/NzbDrone.Core.Test/PerformerTests/RefreshPerformerServiceRescanFixture.cs
@@ -1,0 +1,80 @@
+using System.Collections.Generic;
+using System.Linq;
+using FizzWare.NBuilder;
+using Moq;
+using NUnit.Framework;
+using NzbDrone.Core.MetadataSource;
+using NzbDrone.Core.Movies;
+using NzbDrone.Core.Movies.Performers;
+using NzbDrone.Core.Movies.Performers.Commands;
+using NzbDrone.Core.Test.Framework;
+
+namespace NzbDrone.Core.Test.PerformerTests
+{
+    [TestFixture]
+    public class RefreshPerformerServiceRescanFixture : CoreTest<RefreshPerformerService>
+    {
+        private Performer _performer;
+
+        [SetUp]
+        public void Setup()
+        {
+            _performer = Builder<Performer>.CreateNew()
+                .With(p => p.Id = 1)
+                .With(p => p.ForeignId = "performer-1")
+                .With(p => p.Name = "Performer One")
+                .With(p => p.Monitored = true)
+                .Build();
+
+            Mocker.GetMock<IPerformerService>()
+                .Setup(s => s.GetById(_performer.Id))
+                .Returns(_performer);
+
+            Mocker.GetMock<IProvideMovieInfo>()
+                .Setup(s => s.GetPerformerInfo(_performer.ForeignId))
+                .Returns(_performer);
+
+            Mocker.GetMock<IProvideMovieInfo>()
+                .Setup(s => s.GetPerformerWorks(_performer.ForeignId))
+                .Returns((new List<string>(), new List<string>(), new List<int>()));
+        }
+
+        [Test]
+        public void should_rescan_movies_with_a_single_batch_lookup_and_distinct_ids()
+        {
+            var movies = Builder<Movie>.CreateListOfSize(2).Build().ToList();
+
+            // The performer credit join can return the same movie twice
+            var items = new List<Movie> { movies[0], movies[0], movies[1] };
+
+            Mocker.GetMock<IMovieService>()
+                .Setup(s => s.GetByPerformerForeignId(_performer.ForeignId))
+                .Returns(items);
+
+            Mocker.GetMock<IMovieService>()
+                .Setup(s => s.GetMovies(It.IsAny<IEnumerable<int>>()))
+                .Returns(movies);
+
+            Subject.Execute(new RefreshPerformersCommand(new List<int> { _performer.Id }));
+
+            Mocker.GetMock<IMovieService>()
+                .Verify(s => s.GetMovies(It.Is<IEnumerable<int>>(ids => ids.Count() == 2 && ids.Contains(movies[0].Id) && ids.Contains(movies[1].Id))), Times.Once());
+
+            Mocker.GetMock<IMovieService>()
+                .Verify(s => s.GetMovie(It.IsAny<int>()), Times.Never());
+        }
+
+        [Test]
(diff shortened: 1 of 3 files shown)


plz12345 (changes requested):
Nice, clean change, and the behavioural analysis in the description is accurate — I verified each claim against the code rather than taking it on trust. Two things I'd like addressed before merge, neither of them a correctness bug.

## Should-fix

**1. `RefreshStudioService` gets the identical change with zero test coverage**

`RefreshStudioService.Execute` receives the exact same `Distinct()` + `GetMovies` rewrite as the performer service, but the only new fixture is `RefreshPerformerServiceRescanFixture`. The dedup argument is actually *weaker* for studios — `GetByStudioForeignId` joins only `MovieMetadata`, not `Credit`, so it can't return duplicates today — which means the `Distinct()` there is defending against nothing observable and nothing pins the batch behaviour.

A mirror `RefreshStudioServiceRescanFixture` (the `GetMovies` once / `GetMovie` never / empty-items assertions) should be ~30 lines given `StudioTests/RefreshStudioServiceFixture.cs` already exists as a template. As it stands, half the changed production code has no test.

**2. The batch fetch is unbounded, and this repo chunks at 1000 everywhere else**

`RefreshPerformerService.cs:350` and `RefreshStudioService.cs:303` pass every id in one call.

The parameter-limit angle is a non-issue — `WhereBuilderSqlite.ParseEnumerableContains` explicitly inlines int lists as literals (`// hardcode the integer list if it exists to bypass parameter limit`), same in the Postgres builder. So this is a SQL-statement-size and peak-memory concern, not a crash: a studio with ~30k scenes now builds a single `IN (...)` clause with 30k inlined integers and materialises 30k `Movie` + joined `MovieMetadata` objects at once. Before, peak was one movie.

The precedent is right next door — `RefreshMovieService.cs:410` does `AllMovieIdsOrderByLastInfoSync().Chunk(1000)` then `GetMovies(movieList)` for exactly this reason, and `Execute` in this very file already chunks performer ids at 1000. Wrapping the ids in `.Chunk(1000)` matches the house pattern and keeps all the win.

## Nits

3. **`Fix:` prefix on a change with no behavioural fix.** The description is explicit that nothing user-visible changes ("this is readability rather than a fix"). `Chore:` or `refactor:` describes it better. Not a gate failure — the repo tolerates both prefix styles.

4. **Batch-throw blast radius.** Worth stating plainly for the record: neither `ApplicationException` nor the old `ModelNotFoundException` is caught in `Execute`, so both abort the entire refresh command and skip every remaining performer/studio. The change makes it fail *earlier* (zero rescans instead of partial), which is arguably cleaner. Not a regression, and hardening it belongs in its own PR.

5. **Second fixture rather than extending the existing one.** `PerformerTests/RefreshPerformerServiceFixture.cs` already exists with comparable setup. Separate `...RescanFixture` is defensible for isolation, just flagging the duplication.

## Things I checked that are fine

- **`.Distinct()` is load-bearing, as claimed.** `BasicRepository.Get(ids)` throws when `result.Count != ids.Count()`, and `GetByPerformerForeignId` joins `Credit`, so a two-credit performer genuinely returned the same movie twice. Correct call.
- **TMDB exclusion rewrite is semantically identical.** `int.TryParse` sets the out param to `0` on failure, so the old shared-closure `Select`/`Select`/`Where(!= 0)` and the new `? tmdbId : 0` + `Where(!= 0)` produce the same set.
- **`ToHashSet()` doesn't change comparison semantics.** `List<string>.Contains` and `Enumerable.Contains` both use `EqualityComparer<string>.Default` (ordinal), same as the default `HashSet<string>`. No case-sensitivity shift.
- **Eager materialisation of `scenesToAdd`/`moviesToAdd` is safe.** `AllMovieStashIds()`/`AllMovieTmdbIds()`/`AllMovieTpdbIds()` all return materialised `List<T>` snapshots, so the old deferred re-enumeration for `.Count()` after `AddMovies` already returned the pre-add count. Log values unchanged.
- **The per-entity `AllMovieStashIds()` re-read is correctly preserved** — `AddMovies` runs inside that loop, as the description says.
- `movieIds.Count > 0` is redundant (`Get` already short-circuits on empty), but harmless and the test pins it.

## Local verification

Ran the touched areas on the checked-out branch:

```
  --filter "FullyQualifiedName~PerformerTests|FullyQualifiedName~StudioTests&TestCategory!=IntegrationTest&..."

Passed! - Failed: 0, Passed: 8, Skipped: 0, Total: 8, Duration: 145 ms
```

Build was clean, which also confirms the dropped `using DryIoc.ImTools;` was genuinely unused.

## Checklist

| Gate | Result |
|---|---|
| Head branch meaningfully named | pass |
| Commit prefix | pass (see nit 3) |
| One change per PR | pass — splitting #447 into #448–#451 was the right instinct |
| DB migration note | pass — `NO`, no migration in the diff |
| Tests present | partial — performer covered and passing, studio uncovered |
| Translation keys / `en.json` | n/a, no new user-facing strings |
| ~80% new-code coverage | partial — roughly half the changed production code untested |
| Parser touched | n/a |
| CI | `triage` passed; "Approve fork analysis" pending maintainer approval, so `build_v3` and SonarQube have not run yet |

One note on CI: the workflow hold is the standard fork-PR gate, not a fault in this PR. Someone should approve the run before merge so `unit_test`, `unit_test_postgres` and Sonar actually report — my local run only covered two fixture directories.

PR #451, opened 2026-08-16
Fix: Batch credit loading when matching releases to a studio

ghost:
#### Database Migration

NO

#### Description

When a release has no parsable date, matching falls back to comparing it against every movie in the studio. `MatchMovies` loaded credits inside that loop:

```csharp
if (!movie.MovieMetadata.Value.Credits.Any())
    movie.MovieMetadata.Value.Credits = _creditService.GetAllCreditsForMovieMetadata(movie.MovieMetadata.Value.Id);
```

The guard never memoises anything, because the candidate list is re-read from the DB on each call and `Credits` starts empty, so this fires for effectively every candidate. On a large studio that is thousands of joined queries per release, times every studio the title fuzzy-matches.

`PreloadCredits` now resolves them in one query via new `GetAllCreditsForMovieMetadataIds` and `FindByMovieMetadataIds`. The credit to performer mapper was extracted into `MapPerformerOntoCredit`. `GetPerformerMovies` has a different mapper and is untouched.

`FuzzyMatchReleaseTokens` is unchanged, since its per-candidate query is the last-resort path.

Notes:

- Credits load for all candidates up front rather than lazily for those surviving the earlier checks. Same match results, slightly more rows when candidates match early.
- Movies sharing a metadata id each get their own list copy.
- Removed a duplicated `FindByStudioAndDate`. The only code between the two identical calls is the fuzzy pass, which is read-only.
- Hoisted a loop-invariant `StripSpaces` and materialised two deferred LINQ chains that were re-running normalisation per credit.
- Empty and null input are guarded, since SQLite `IN ()` is a syntax error. Large integer lists are fine, as both the SQLite and Postgres builders inline them.

N/A

#### Todos

- [x] Tests: `PreloadCreditsFixture` (batch called once, singular never called, not called when credits are already present, shared-metadata isolation), `CreditServiceFixture` (grouping, missing ids, empty input skips the repo), `CreditRepositoryFixture` (`DbTest` on SQLite, including a 2,000-id list).
- [x] Translation Keys: N/A

- N/A

### DIFF
--- a/src/NzbDrone.Core.Test/MovieTests/CreditTests/CreditRepositoryFixture.cs
+++ b/src/NzbDrone.Core.Test/MovieTests/CreditTests/CreditRepositoryFixture.cs
@@ -1,0 +1,61 @@
+using System.Collections.Generic;
+using System.Linq;
+using FluentAssertions;
+using NUnit.Framework;
+using NzbDrone.Core.Movies.Credits;
+using NzbDrone.Core.Test.Framework;
+
+namespace NzbDrone.Core.Test.MovieTests.CreditTests
+{
+    [TestFixture]
+    public class CreditRepositoryFixture : DbTest<CreditRepository, Credit>
+    {
+        private static Credit GivenCredit(int movieMetadataId, string name)
+        {
+            return new Credit
+            {
+                MovieMetadataId = movieMetadataId,
+                PersonName = name,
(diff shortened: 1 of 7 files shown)


plz12345:
What is your actual issue here?  Yes, I get that it _looks_ inefficient, but based on the shoddy previous PR's cache implementation, it's clear there is a lack of understanding in the code base.  What observed-in-real-world problem are you solving with this?

PR #533, opened 2026-08-27
Convert restart and shutdown off Redux

plz12345:
#### Database Migration

NO

#### Description

Prerequisite for section 6b (General settings), split out the way #516 split the
pending stores out of section 1. General is the only page that restarts the app
from a Save, so it cannot come off `react-redux` while `restart` is a thunk.

`System/useSystem.ts` replaces both `systemActions` thunks. Neither ever wrote
to a reducer — the slice's `defaultState` has been `{}` since status, health,
tasks, backups, updates and logs became queries — so `restart` is a mutation
whose `onSuccess` does exactly what the thunk's `.done()` did: set the app
store's `isRestarting` and start the ping loop that clears it.

That empties `systemActions.js`, so it goes, and `Store/Actions/index.js` is
down to **one** registered slice (`settings`). `react-redux` 80 → 78.

`GeneralSettingsConnector` takes `restart` as an own prop from its function
wrapper for now; the connector goes in 6b.

Verified on the running instance: **Restart** from the header menu sends
`POST /api/v3/system/restart`, which answers 200, and the app goes down and
comes back reconnected. Probed separately to confirm the response resolves
rather than being cut off by the shutdown — that is what `onSuccess` hangs on.
**Shutdown was not exercised**, for the obvious reason.

Stacked on #532; base retargets to `eros-develop` when that merges.

#### Todos

- [ ] Tests

n/a


plz12345:
Reopening under a branch name that matches the change; the General settings work needs redux/settings-general.

PR #594, opened 2026-09-01
New: allow unmonitored discovery items

anoncod69:
#### Database Migration
YES - 026

#### Description
Adds per-Performer, per-Studio, and per-Collection **Monitor New Items** controls plus a global master switch under **Settings → General**.

Both levels default to enabled, preserving current behavior. Effective monitoring is:

`global master switch && entity Monitor New Items`

When the global switch is disabled, metadata and RSS discovery continue normally and newly discovered works are still added to the library, but they are added unmonitored and Search-on-Add is suppressed. Interactive/manual search and manual grabs remain available. When the global switch is enabled, each entity's own setting applies.

Migration 026 persists the three entity-level settings for existing databases. The global setting uses Whisparr's existing configuration store and needs no additional migration. Existing works, Import Lists, download clients, indexers, RSS settings, and entity monitoring semantics are unchanged.

The UI changes use standard checkbox controls following the existing Media Management and entity add/edit form patterns; no layout or visual-system changes.

#### Todos
- [x] Tests

#### Verification
- 4,018 non-DB Core tests passed; 26 skipped
- 168 DB tests passed
- 54 API tests passed
- 44 focused config/refresh tests passed
- Full solution build passed with 0 warnings and 0 errors
- Frontend ESLint and Prettier passed
- Webpack build passed
- Exact combined diff credential scan: 0 findings

- Fixes whisparr/whisparr#1129


plz12345:
Question on intended semantics before this lands — Whisparr/Whisparr-Eros#850's text and the implementation here agree with each other, but I want to make sure they match product intent.

**As implemented**, MonitorNewItems is a strict AND gate *on top of* existing entity monitoring:

- Sync guard (RefreshPerformerService.cs:154 / RefreshStudioService.cs:104): `if (!Monitored && !MoviesMonitored) return;` — so on an entity that's **fully unmonitored, MonitorNewItems alone does nothing**. No catalog sync, no new works, regardless of the flag.
- To get a performer's works into the library with *nothing* monitored today, you'd enable Monitored or MoviesMonitored while leaving MonitorNewItems off — works then arrive unmonitored. Works, but via two toggles where users will expect one.

If instead the intent is "pull in a performer's works **without** monitoring the performer" (i.e., MonitorNewItems alone enables discovery), this PR doesn't do that and would need:

1. Sync guards to include `|| entity.MonitorNewItems` — and the same for PerformerAddedHandler / StudioAddedHandler, which currently only refresh-on-add when Monitored/MoviesMonitored is set.
2. **Explicitly** setting MonitorNewItems=false at auto-creation (RefreshMovieService.cs ~L165/213, where performers/studios get hardcoded `Monitored = false`). This isn't optional under that design: the model default is `= true` (Performer.cs / Studio.cs), and BasicRepository writes every property on insert — so with an OR gate plus the current default, *every scene added would fan out into all its performers'/studios' full catalogs as unmonitored works*.
3. New works arriving via that path defaulting their Monitored state from the entity's monitor flags (unmonitored entity → unmonitored works).

The auto-added-entities behavior is otherwise safe under either design: they come up fully unmonitored (`Monitored = false` hardcoded, `MoviesMonitored = false`), and all three refresh services early-return on them, so there's no cascade as written.

Happy to discuss which semantics we're standardizing on — the issue body says "the entity itself remains monitored", but if the OR-gate version is what users actually want, let me know before I merge-review this.

anoncod69:
Thanks for calling this out. The strict AND semantics are intentional and match the product intent in Whisparr/Whisparr-Eros#850.

`MonitorNewItems` is not intended to enroll an otherwise fully unmonitored Performer, Studio, or Collection in discovery. The entity must already be eligible for refresh through its existing `Monitored` / `MoviesMonitored` state. This new flag controls only whether works discovered through that existing refresh are added as monitored. With the flag off, an eligible entity still refreshes and adds newly discovered works, but those works are added unmonitored.

The global master switch is an additional AND gate over the entity setting. We therefore do not want the OR-gate behavior or the resulting auto-created-entity catalog fan-out.

I pushed `5525dcd58ab3dd3a9fa7e90d3153287df6bba5f4` to make this relationship explicit in the UI: all entity-level controls now state that the entity must already be monitored for discovery, and the global help text states that it does not activate discovery for a fully unmonitored entity. No sync guard, added-handler, default, or backend semantic changed.

plz12345:
Ah, gotcha, so it covers the "I want all this performer's works in library, but I want to pick which ones I download", and same for studio.  Noted, and agree on logic now.

plz12345 (changes requested):
Thanks for this — the feature is well built and the test coverage is genuinely thorough. Two things need to settle before it can merge, both about naming and placement rather than behaviour.

## Blocking

### 1. Rename to the `Whisparr*` prefix

The PR introduces `EnableNewItemMonitoring` (global) and `MonitorNewItems` (per-entity), neither prefixed. This project prefixes its own settings deliberately, to keep them isolated from upstream Radarr/Sonarr code — every member of the `Whisparr` fieldset in `HostConfigResource.cs` follows that rule. `MonitorNewItems` is also a name Sonarr already uses upstream, for a differently-typed field (`Series.MonitorNewItems`, a `NewItemMonitorTypes` enum rather than a bool).

Proposal: **`WhisparrMonitorNewItems`** for both, keeping the global switch and the entity-level property matched.

This is blocking rather than a follow-up because of `026_add_monitor_new_items.cs` — the column name lands in three tables the moment this merges, and renaming afterwards costs a second migration against a released schema. The config key string in `ConfigService.cs` has the same property: free to change now, since nothing has persisted a value yet, but a `Config`-table data migration later.

The knock-on is the whole PR — all 38 files carry one of these identifiers. The spots that aren't purely mechanical:

- the migration's three column names, and the `SELECT` statements in its fixture
- the `Config`-table key string in `ConfigService.GetValueBoolean`/`SetValue`
- the three `en.json` keys (`MonitorNewItems`, `MonitorNewItemsEntityHelpText`, `MonitorNewItemsHelpText`), which should follow the existing `WhisparrCacheMovieAPI` / `WhisparrCacheMovieAPIHelpText` pattern
- the two zustand defaults stores — a renamed key in an already-persisted `addPerformerDefaults` / `addStudioDefaults` blob falls back to the new default rather than migrating. Harmless here since both default to `true`, just worth knowing it isn't a migration.

### 2. Move the global switch to Settings → General → Whisparr, non-advanced

It currently renders in the **File Management** fieldset of `MediaManagement.tsx`. Two reasons to move it: it's a monitoring control rather than file handling, and `MediaManagement.tsx` tracks upstream closely, so an insertion there is a standing merge-conflict source. The `Whisparr` fieldset in `WhisparrSettings.tsx` is already Eros-only and is where this project's own toggles live.

Concretely:

- move `EnableNewItemMonitoring` from `MediaManagementConfigResource.cs` to `HostConfigResource.cs` — both map off `IConfigService`, so `ConfigService.cs` itself is unchanged, though the `// Permissions (Media Management)` grouping comment in `IConfigService.cs` should follow it
- swap the frontend typing from `typings/Settings/MediaManagement.ts` to `typings/Settings/General.ts`
- render the `FormGroup` in `WhisparrSettings.tsx` **without** `isAdvanced`
- rename `MediaManagementConfigResourceFixture` to a `HostConfigResourceFixture` — its reflection test over `IConfigService` still holds as written

One knock-on worth flagging: every current member of that fieldset is `isAdvanced={true}`, so a non-advanced entry makes the `Whisparr` fieldset visible to users who don't have advanced settings turned on. That's the intent, but it is a visible change.

Folding this into the same pass as the rename keeps it to one round-trip.

## Should-fix

### Resource properties default to `false`, so a client that omits the field silently disables monitoring

`PerformerResource.cs`, `StudioResource.cs`, `CollectionResource.cs`.

The Core models correctly carry `= true` initialisers, but the resources don't. `ToModel()` copies `resource.MonitorNewItems` unconditionally and `ApplyChanges` then overwrites the stored value. Failure scenario: a third-party tool `PUT`s a performer body built from a pre-594 shape, with no `monitorNewItems` key. System.Text.Json leaves the property at its `false` default, and that performer quietly stops monitoring newly discovered scenes with no user action.

One character each: `public bool MonitorNewItems { get; set; } = true;`. The UI is unaffected — it spreads the full fetched object.

### Bulk editors can't set the new flag

`PerformerEditorResource.cs` and `StudioEditorResource.cs` both carry `bool? MoviesMonitored` and are wired to live UI (`useEditPerformersModalMutation`, `useEditStudiosModalMutation`), so users can mass-edit monitoring but have to open each entity individually for this one. A feature gap rather than a bug — fine as a follow-up.

### The global switch doesn't cover import lists

`MonitorNewItemsHelpText` calls it a "Global master switch for monitoring newly discovered works", but `ImportListSyncService.cs` is untouched — list-added items stay monitored per list settings. The second sentence does scope it to "entities that are already monitored for discovery", so it's defensible, but "Global master switch" oversells it.

## Nits

- `AddNewStudioModalContent.tsx` — `errors={settings.monitorNewItems?.errors}` after `{...settings.monitorNewItems}` is redundant, since the spread already carries `errors`, and it's inconsistent with the other three call sites.
- Commit 3 uses `docs:`, outside both the Servarr (`New:`/`Fix:`/`Chore:`) and conventional sets. Irrelevant once squashed — the PR title is fine.
- The PR adds `ExceptionVerification.ExpectedErrors`/`ExpectedWarns` to three pre-existing tests in `RefreshPerformerServiceFixture`. Those assertions only bite in Debug builds, since `LoggingTest.LoggingDownBase` gates on `BuildInfo.IsDebug`. They're a genuine strengthening, just unrelated to the feature.

## What I verified

Read the changed files in full rather than only the hunks. The gating logic is applied at all nine discovery add sites — three metadata sources each for performer and studio, plus the collection path. I grepped `AddMovieMethod.` and `Monitored = true` across Core to confirm nothing was missed; the only remaining `Monitored = true` is in `IdentificationService`, which is manual import. `MonitorNewItems` is carried through `ApplyChanges` on all three models and through the performer merge path in `RefreshPerformerService`. Migration 026 is the next free number in the Eros range. Zustand's `persist` shallow-merges, so existing users' saved add-defaults pick up `monitorNewItems: true` rather than `undefined`. Both levels default to `true`, so behaviour is unchanged for anyone who doesn't touch the setting — that part of the design is sound.

Local test run on the PR branch:

- `add_monitor_new_itemsFixture` + `RefreshCollection`/`Performer`/`StudioServiceFixture` + `ConfigServiceFixture` — 47 passed, 0 failed
- `NzbDrone.Api.Test` `~ResourceFixture` — 19 passed, 0 failed
- migration fixture on its own — 3/3, performers, studios and collections all defaulting to `true`

## Checklist

| Gate | Result |
| --- | --- |
| Head branch named meaningfully | pass — `feat/monitor-new-items` |
| Commit prefixes | `docs:` on commit 3; harmless under squash |
| One feature per PR | pass — cohesive |
| DB migration note | pass — "YES - 026", matches the added migration |
| Tests checkbox and tests present | pass — 8 new/changed fixtures |
| Translation keys | pass — 3 keys, all used, correctly sorted (names need the prefix per blocker 1) |
| ~80% new-code coverage | pass — every branch of the global/entity matrix is covered |
| CI | pass — backend x10, frontend, unit_test x3, unit_test_postgres x3, integration_test x3, CodeQL |
| SonarQube | "Approve fork analysis" pending — fork-PR workflow approval, not a fault in the PR |

`mergeStateStatus: BLOCKED` with `mergeable: MERGEABLE` is just the pending review. Note `eros-develop` has moved on since you branched (#595 merged); still mergeable.

anoncod69:
Thanks for the thorough review — all items are addressed and pushed as `3f6af6433` (built on your `eros-develop` merge `0c69d228`, so your sync commit is preserved).

**Blocking**

1. **Rename to `Whisparr*` prefix** — Done. `EnableNewItemMonitoring` and `MonitorNewItems` are now `WhisparrMonitorNewItems` across all 41 files: models, resources, refresh services, migration 026 columns (performers/studios/collections) and its fixture `SELECT`s, the `Config`-table key, the three `en.json` keys (now `WhisparrMonitorNewItems`, `WhisparrMonitorNewItemsEntityHelpText`, `WhisparrMonitorNewItemsHelpText`, alphabetically ordered per the existing `WhisparrCacheMovieAPI` pattern), the zustand defaults stores, and all UI call sites. A repo-wide grep confirms zero bare `MonitorNewItems` / `EnableNewItemMonitoring` identifiers remain.

2. **Global toggle moved to Settings → General → Whisparr, non-advanced** — Done. Property moved from `MediaManagementConfigResource.cs` to `HostConfigResource.cs` (still maps off `IConfigService`; grouping comment updated), frontend typing moved from `MediaManagement.ts` to `General.ts`, FormGroup removed from the File Management fieldset in `MediaManagement.tsx` and rendered in `WhisparrSettings.tsx` without `isAdvanced`. `MediaManagementConfigResourceFixture` renamed to `HostConfigResourceFixture`.

**Should-fix**

3. **Resource defaults** — `PerformerResource`, `StudioResource`, `CollectionResource` now all have `= true` on `WhisparrMonitorNewItems`, so a client omitting the field can no longer silently disable monitoring.

4. **Help-text scoping** — The global help text no longer claims full "global master switch" coverage; it now states it controls monitoring of newly discovered works from Performer/Studio/Collection refresh and explicitly does not affect Import List behavior.

**Nit**

5. The redundant `errors` prop after the spread in `AddNewStudioModalContent.tsx` is removed.

**Verification on the merged head**

- Solution build: 0 warnings / 0 errors
- Migration + Refresh services + ConfigService fixtures: 44 passed, 0 failed
- API `~ResourceFixture`: 19 passed, 0 failed
- ESLint + Prettier on all changed frontend files and `en.json`: passed
- Frontend webpack build: passed
- `git diff --check`: clean

Bulk-editor support and Import List coverage remain accepted follow-ups per your notes.

plz12345 (approved):

PR #644, opened 2026-09-04
Fix: bound media cover processing during refresh

anoncod69:
#### Database Migration

NO

#### Description

Bounds media-cover processing triggered by `MovieUpdatedEvent` so a large movie refresh cannot create an unrestricted number of asynchronous cover tasks and retain movie object graphs in proportion to the refresh backlog.

The change:

- adds a capacity-100 queue keyed by movie ID;
- coalesces duplicate pending updates while preserving the latest handler-entry value;
- applies producer backpressure only to new unique IDs when capacity is exhausted;
- processes work with a fixed worker count aligned with the existing CPU-based resize limit;
- preserves one `MediaCoversUpdatedEvent` per processed queue entry;
- isolates per-item failures so one cover failure does not terminate a worker;
- abandons queued covers during shutdown and enforces one five-second overall deadline for active workers;
- cancels blocked producers and rejects delayed producers once shutdown begins;
- reconciles late worker exits without creating a restart-only lifecycle.

The queue is owned by `MediaCoverService`; the global event aggregator is not throttled, avoiding nested-publication deadlock risk. Performer and studio cover handlers are unchanged.

Because `MovieUpdatedEvent` synchronous handlers run inline, full-queue backpressure can pause the `RefreshMovieCommand` publisher while cover capacity drains. The current command/event contract does not carry the UI command-cancellation token to this handler; adding that propagation requires broader command infrastructure changes and is intentionally deferred from this two-file fix. Application shutdown still cancels blocked producers.

Regression coverage includes a deterministic 10,000-update duplicate-heavy stream, capacity and fixed-concurrency assertions, reversed waiter ordering, normal multiple-overflow release, unique-update delivery, completion events, exception isolation, shutdown abandonment, bounded stuck shutdown, delayed-producer rejection, and start/shutdown races.

Post-sync verification against current `eros-develop` (`0dc1494b7c14bb8ab265fef51863479e31b3417d`):

- focused `MediaCoverServiceFixture`: 28 passed, 0 failed;
- stale-producer regression: 30 consecutive passes under forced ordering;
- complete `Whisparr.Core.Test`: 4,438 passed, 26 skipped, 0 failed;
- `git diff --check`: clean;
- exact two-file diff credential scan: 0 findings.

Not applicable; backend-only change.

#### Todos

- [x] Tests
- [x] Translation Keys (not applicable; no user-facing strings)

- Fixes whisparr/whisparr#1246


plz12345 (changes requested):
Nice piece of work overall — the diagnosis in whisparr/whisparr#1246 is thorough and the fixture is far more rigorous than most concurrency tests we get. Requesting changes on one real race that the tests themselves caught, plus two behavioural regressions from the sync-handler conversion that I think need a decision before this merges.

## Blocking

### 1. Coalescing is not last-writer-wins once an entry has been dequeued

`src/NzbDrone.Core/MediaCover/MediaCoverService.cs:744`

`Handle` compares `sequence` only against what is currently in `_pendingMovieCovers`. The worker removes the entry at `MediaCoverService.cs:853` *before* processing it, so the map no longer remembers that a newer sequence for that movie was already handled. There is no last-processed-sequence guard.

Failure scenario: the queue is full, and producers for movie 12000 with seq=N ("old") and seq=N+2 ("latest") are both blocked on `_movieCoverQueueSlots`. `SemaphoreSlim` release order is not guaranteed FIFO, so "latest" can wake first, enqueue, and be dequeued and removed by a worker immediately. "old" then wakes, finds 12000 *not* pending, adds it, and it is processed second. The movie is processed twice and the **final** `MediaCoversUpdatedEvent` — and the cover state written to disk — comes from the *stale* snapshot.

This is what CI hit, at `src/NzbDrone.Core.Test/MediaCoverTests/MediaCoverServiceFixture.cs:596`:

```
Expected string to be a match with the expectation, but it differs at index 39:
  "…duplicate-old"      (actual)
  "…duplicate-latest"   (expected)

Error processing media covers for movie 12000
  System.InvalidOperationException: Invalid attempt made to decrement the event's count below zero.
   at System.Threading.CountdownEvent.Signal()
   at MediaCoverServiceFixture...b__5(MediaCoversUpdatedEvent message):line 552
   at NzbDrone.Core.MediaCover.MediaCoverService.ProcessMovieCoverQueue():line 861
```

The 105th signal on a `CountdownEvent(104)` is movie 12000 completing twice. So "preserves one `MediaCoversUpdatedEvent` per processed queue entry" does hold — but the fixture asserts one event *per movie*, which the implementation does not guarantee.

Please fix the implementation rather than the assertion. The stale cover surviving is the actual defect; the duplicate event is just the symptom. Tracking a per-movie last-processed sequence and skipping a lower one should close it.

### 2. CI is red

`unit_test_postgres (17)` FAILURE on the run for dbcb21f. Everything else is green: all 10 `backend` targets, `frontend`, `unit_test` x3, `unit_test_postgres` 16 and 18, `integration_test` x3, CodeQL x3. No SonarQube "Build and Analyze" ran (fork PR — `Approve fork analysis` is pending, which is expected and not a fault of this PR).

Worth stressing that this is not the periodic flake that job sometimes shows: the failing test is one this PR adds, and the trace lands inside the new `ProcessMovieCoverQueue`. It passed on postgres 16 and 18 and on all three `unit_test` runners, so it is timing-dependent — which makes it a permanent flake source until finding 1 is fixed.

I ran `MediaCoverTests` locally on the branch: 30/30 pass, and I could not reproduce the race in 7 runs (36 executions of the failing test, `DOTNET_PROCESSOR_COUNT=4`, including 6 concurrent runs for contention). It needs CI-level scheduling pressure. Local green does not clear it.

## Should-fix

### 3. The `IHandle` conversion blocks the refresh command thread, and a UI cancel will not unblock it

`src/NzbDrone.Core/MediaCover/MediaCoverService.cs:36`

`EventAggregator.PublishEvent` invokes sync handlers inline on the publisher's thread, in order (`src/NzbDrone.Core/Messaging/Events/EventAggregator.cs:102-119`). The publisher is `RefreshMovieService.cs:269`, so at queue capacity the `RefreshMovie` command now blocks on cover downloads. That is the intended backpressure — but `_movieCoverQueueSlots.Wait(intakeCancellation)` is cancellable *only* by app shutdown, so cancelling the refresh from the UI leaves the thread parked until covers drain. Consider threading the command's cancellation token through. It also serialises ahead of `MovieController.Handle(MovieUpdatedEvent)` (`src/Whisparr.Api.V3/Movies/MovieController.cs:989`), delaying SignalR movie updates by the same amount.

### 4. Draining on shutdown adds up to 30s to every stop/restart

`src/NzbDrone.Core/MediaCover/MediaCoverService.cs:795-812`

Shutdown releases the workers and joins them while they drain up to 100 queued covers, capped by `MovieCoverShutdownTimeout` (default 30s). `AppLifetime.Handle` is `EventHandleOrder.Last`, so this runs first and delays the real shutdown. Docker's default SIGTERM grace is 10s, so a restart mid-refresh gets SIGKILL'd anyway. Covers are re-fetched on the next refresh — dropping the queue and cancelling in-flight work, or a much shorter deadline (~5s), seems the better trade.

### 5. Test-only hooks in production code

`MovieCoverQueueSlotAcquired` (`MediaCoverService.cs:53`, invoked at `:723`), the settable `MovieCoverShutdownTimeout`, and five `internal` observability properties exist purely for the fixture. The callback in particular is an injected interleaving point on the hot path. `InternalsVisibleTo` makes it compile, but a `MediaCoverQueueOptions`-style seam would keep the service clean.

### 6. Complexity relative to the problem

317 lines of hand-rolled lifecycle (four-state machine, generation counters, timed-out-shutdown reconciliation, late-worker recovery) to bound a work queue. The generation/restart machinery is unreachable in production — `ApplicationStartedEvent` is published exactly once, from `src/NzbDrone.Host/AppLifetime.cs:68` — so it exists only to satisfy tests. A `System.Threading.Channels` bounded channel plus a dedupe dictionary would cover the issue's stated requirements in a fraction of the surface. Not a merge blocker on its own, but it is a lot of concurrency for the project to own, and it is the kind of surface that hid finding 1.

## Nits

- `MediaCoverService.cs:690`: "because the media cover queue is stopping" also fires when the state is `Stopped` (not yet started) or `ShutdownTimedOut`. Worth naming which.
- `MediaCoverService.cs:761`: the replaced `CancellationTokenSource` is never disposed, and neither semaphore is. Trivial in practice.
- The new `internal` properties sit above the private field block, inverting the file's existing fields-then-members ordering. Analyzers did not complain.
- whisparr/whisparr#1246 is already CLOSED and has no labels — it looks auto-closed for being labelless rather than resolved. Worth relabelling/reopening so the `Fixes` link does something on merge.

## Checklist

| Gate | Result |
| --- | --- |
| Head branch named meaningfully | pass |
| Commit prefix (`fix:`) | pass |
| One change per PR | pass |
| DB migration note | pass — "NO", no migration in diff |
| Tests / Translation Keys checkboxes | pass |
| Linked issue | pass (issue state questionable, see nits) |
| `en.json` keys | n/a — log messages only, no user-facing strings |
| NUnit coverage >=80% new code | pass — 844 test lines against 317 service lines |
| CI status | **fail — `unit_test_postgres (17)`**; all other gates green, no SonarQube run (fork PR) |
| Local verification | `MediaCoverTests` 30/30 pass x7 runs; race not reproducible locally |

`mergeStateStatus: BLOCKED` with `mergeable: MERGEABLE` is the failing check plus awaiting approval, not a conflict.

anoncod69:
Addressed the review against the current `eros-develop` head.

1. **Stale blocked producer / last-writer-wins** — fixed. The service now keeps a per-movie admitted-sequence high-water mark for its lifetime. Both the initial intake path and the post-slot-acquisition path reject an older producer even after newer work has been dequeued. `should_skip_an_older_producer_after_the_latest_update_has_been_dequeued` forces that ordering; it passed 30 consecutive contention runs.

2. **PostgreSQL 17 CI failure** — the failing assertion and 105th completion signal were symptoms of the same race above, not an assertion problem. The implementation was corrected and the assertion retained. A fresh CI run is now triggered by this push.

3. **Synchronous backpressure and UI cancellation** — the bounded backpressure remains intentional: it prevents the refresh publisher from creating an unbounded retained-work backlog. The current `RefreshMovieCommand` / `IExecute` path does not propagate the UI cancellation token to `MovieUpdatedEvent` handlers, so safely threading UI cancellation through requires changes outside this two-file repair. That limitation and the possible SignalR delay are now explicit in the PR description. Application shutdown still cancels blocked producers.

4. **Shutdown latency** — changed. Shutdown clears queued, not-yet-started covers, cancels blocked producers, and uses one overall 5-second deadline for active workers rather than attempting to drain up to 100 covers for 30 seconds.

5. **Test-only hooks** — consolidated under a single internal `MovieCoverQueueTest` seam. The interleaving callback and observability access no longer sit as loose service members.

6. **Lifecycle complexity** — reduced. Generation counters, restart-only state, and late-restart recovery were removed. The remaining queue/dictionary coordination is needed for keyed coalescing, unique-ID capacity, and the persistent sequence high-water; retaining a pending entry through processing would change capacity semantics and could lose an update arriving during active processing.

Nits:

- Lifecycle state is now only `Running`, `Stopping`, and `Stopped`; no `ShutdownTimedOut` or generation state remains.
- The replaced per-generation cancellation-source path was removed. The remaining synchronization primitives are service-lifetime members of the singleton.
- Internal test controls were moved to the consolidated seam at the end of the class.
- whisparr/whisparr#1246 remains closed by repository automation. Its body was corrected to the issue template and a maintainer review/reopen request was posted; the contributor account cannot reopen it.

Verification on final head `e7a799f7278f897766e3a8565171038989116685` against base `0dc1494b7c14bb8ab265fef51863479e31b3417d`:

- `MediaCoverServiceFixture`: 28 passed, 0 failed
- stale-producer regression: 30 consecutive passes under forced ordering
- complete `Whisparr.Core.Test`: 4,438 passed, 26 skipped, 0 failed
- `git diff --check`: clean
- changed paths: exactly the service and its fixture
- exact diff credential scan: 0 findings

An isolated Claude Opus review found no blocking issue with the high-water repair and agreed that end-to-end UI cancellation is outside the safe two-file scope.

anoncod69:
Follow-up: the fresh `unit_test_postgres (17)` run exposed one remaining ordering window in `should_release_multiple_overflow_producers_and_recover_capacity_during_normal_processing`.

The first repair retained high-water state after dequeue, but it did not publish a unique movie’s sequence until after that producer acquired queue capacity. Therefore an older blocked waiter could acquire the first released slot and complete before the already-waiting latest producer updated the high-water mark.

The follow-up records the newest sequence under the queue lock before waiting for capacity, then uses a strict `<` stale check after acquisition. This means:

- if the old waiter wakes first, it sees the newer pre-registered sequence and releases the slot without enqueueing;
- if the latest waiter wakes first, it matches the high-water sequence and proceeds;
- no assertion or expected completion count was relaxed.

Verification before push:

- exact CI-failing regression: passed;
- same regression: 30 consecutive process-level runs passed;
- `MediaCoverServiceFixture`: 28/28 passed;
- complete `Whisparr.Core.Test`: 4,438 passed, 26 skipped, 0 failed;
- `git diff --check`: clean;
- incremental credential scan: 0 findings.

A second isolated Claude Opus concurrency review examined this exact incremental change and returned **APPROVE**, no blockers, high confidence.

plz12345 (changes requested):
Re-reviewed at `0cce4cd`. Both blocking findings from the last round are genuinely fixed — thanks for going after the implementation rather than the assertion. Re-blocking on one new item: a test in this PR that fails 100% of the time on any machine reporting two or fewer CPUs.

## Previously blocking — resolved

**1. Last-writer-wins after dequeue — fixed.** `_movieCoverHighWaterSequences` (`MediaCoverService.cs:68`) is a per-movie high-water mark retained for the service lifetime, deliberately not cleared on dequeue. Both gates are correct, and the asymmetry between them is the right call:

- intake (`:650`) rejects `sequence <= highWater`;
- post-acquisition (`:695`) rejects `sequence < highWater`, and the sequence is registered at `:666` *before* the blocking `Wait`, so a stale waiter that wakes first sees the newer mark and releases the slot.

I traced every exit from `Handle` — cancellation, not-Running, stale, coalesce-into-pending, enqueue — and the highest sequence for a movie always ends up either enqueued or merged into a pending entry. No update is dropped.

Slot accounting balances too. `_movieCoverQueueSlots` is constructed with `maxCount = 100`, so an over-release would throw `SemaphoreFullException`. It can't: `_pendingMovieCovers.Count == _movieCoverQueue.Count` holds invariantly (add+enqueue at `:713-715`, remove+release under the same lock at `:830-833`), and shutdown's `Release(N)` at `:770` reads `N` under the lock, so producers in flight — which release their own slot at `:691` — are never double-counted.

**2. CI — green.** All 10 `backend` targets, `frontend`, `unit_test` x3, `unit_test_postgres` **16/17/18**, `integration_test` x3, CodeQL x3, `check`, `triage`, all `deploy / package`. `Approve fork analysis` is still pending, so SonarQube "Build and Analyze" has still never run on this PR.

Also resolved: shutdown latency (#4) — now abandons the queue with one 5s overall deadline (`:775-805`); test hooks (#5) — consolidated into `MovieCoverQueueTestSeam`.

## Blocking

### 1. `should_preserve_handler_entry_order_when_full_queue_waiters_resume_in_reverse` fails deterministically with one worker

`src/NzbDrone.Core.Test/MediaCoverTests/MediaCoverServiceFixture.cs:331`

The test reads `MovieCoverWorkerCount` dynamically, but its choreography needs **at least two** slots to free. With one worker, releasing the initial blocker frees exactly one slot; `older` takes it and parks inside the `SlotAcquired` hook still holding it, and `releaseOldWaiter` is only set *after* `newer.Wait(5000)` — which therefore can never succeed. The assertion fails and the test burns 10s instead of 90ms.

Reproduced on this branch, isolated, 100% deterministic (`DOTNET_PROCESSOR_COUNT=2` gives `ceil(2/2) = 1` worker):

```
pc=2 run1: Failed!  - Failed: 1, Passed: 0, Total: 1, Duration: 10 s
pc=2 run2: Failed!  - Failed: 1, Passed: 0, Total: 1, Duration: 10 s
pc=3 run1: Passed!  - Failed: 0, Passed: 1, Total: 1, Duration: 83 ms
pc=4 run1: Passed!  - Failed: 0, Passed: 1, Total: 1, Duration: 87 ms
```

This is test-only — in production there is no hook, so a stale producer releases the slot immediately under the lock and nothing stalls. It stays blocking because it is a guaranteed red build for anyone on a 1-2 vCPU dev box or a `--cpus=2` container, and CI only passes by the accident of runner size. Either `Assume.That(workerCount >= 2)` and skip, or free a second slot before waiting on `newer`.

## Should-fix

### 2. `_movieCoverHighWaterSequences` is never pruned

`src/NzbDrone.Core/MediaCover/MediaCoverService.cs:68`

Retaining the mark past dequeue is exactly what makes the fix above work — that part is right. But entries are only removed at shutdown (`:765`), so the map accumulates one `int -> long` per distinct movie ID for process lifetime, including deleted movies. A few MB at worst, but it is unbounded growth in a PR whose purpose is bounding retention. `HandleAsync(MoviesDeletedEvent)` (`:983`) is the natural place to purge.

### 3. Deleted movies can still have covers written after deletion

`src/NzbDrone.Core/MediaCover/MediaCoverService.cs:983`

`HandleAsync(MoviesDeletedEvent)` deletes the cover folder but does not drop queued entries for those IDs, so a worker can re-create the folder afterwards. The race predates this PR, but the window was previously microseconds — work started immediately — and is now up to 100 queued entries deep. The same purge as #2 closes it.

### 4. Sync backpressure and UI cancel — unchanged, now documented

`src/NzbDrone.Core/MediaCover/MediaCoverService.cs:674`

`_movieCoverQueueSlots.Wait(intakeCancellation)` is still cancellable only by app shutdown, so cancelling a refresh from the UI leaves the publisher thread parked. The PR body now states this as a deferred limitation, which is a fair place to leave it — recording it as accepted rather than as a new objection. Note the publish at `RefreshMovieService.cs:124` sits inside a `catch (MovieNotFoundException)` before a rethrow, so that path can now block too.

### 5. The complexity note did not shrink

Generation counters and the restart-only lifecycle are gone, which is a real improvement. But the service diff went from +317 to **+370** — the high-water map cost more than the lifecycle machinery saved. Still 370 lines of hand-rolled concurrency against a `System.Threading.Channels`-shaped problem. Not a merge blocker, just noting the finding stands.

## Nits (all carried over)

- `:646` "because the media cover queue is stopping" also fires for `NotStarted` and `Stopped`.
- Neither semaphore nor `_movieCoverIntakeCancellation` is disposed. Singleton, so academic.
- `:47-51` the `internal` members still sit above the private field block, inverting the file's ordering.
- `HandleAsync(MovieUpdatedEvent)` (`:967`) is still `public` with the aggregator's conventional name, though the class no longer implements `IHandleAsync<MovieUpdatedEvent>`. Re-adding that interface later would silently double-process.
- `Handle(ApplicationShutdownRequested)` no-ops when state is `NotStarted` (`:754`), so a later `ApplicationStartedEvent` would start workers post-shutdown. Unreachable via `AppLifetime`.
- whisparr/whisparr#1246 is still closed. I will sort out the relabel/reopen from this side so the `Fixes` link does something on merge.

## Local verification

Checked out `0cce4cd`, built, ran `FullyQualifiedName~MediaCoverTests` (32 tests):

- **152 executions at default CPU count** (18 cores, 9 workers): 151 green, **1 failure** on the very first run whose output I did not capture and could not reproduce in 151 further attempts — 30 sequential, 30 at 6-way concurrency, 60 at 12-way oversubscription. I cannot tell you what it was; reporting it because it happened.
- **24 executions at `DOTNET_PROCESSOR_COUNT=2`**: 24 failures, all the test in finding 1.
- **24 executions at `DOTNET_PROCESSOR_COUNT=4`**: all green.

`git diff --check`: clean. Two files, +1261/-1.

anoncod69:
Addressed the latest review feedback in `8e840b9feae602dcbbc31227b8b02d4e008fc36e`.

### Deletion and queue lifecycle
- `MoviesDeletedEvent` now purges matching pending cover work before deleting cover folders.
- Purged queue entries release their capacity slots, while queue wake permits are reconciled without disturbing survivor FIFO order.
- Producers register their per-movie sequence before waiting for capacity. Producer/worker accounting plus temporary deletion tombstones reject continuations that entered before deletion while still allowing genuinely newer post-deletion updates.
- Workers are marked active while holding the queue lock, closing the dequeue/accounting race. If pre-deletion active work recreates a cover folder, worker completion removes it again.
- Idle high-water state is removed on deletion; temporary tombstone/high-water state is cleared after related producers and workers drain, avoiding lifetime per-movie retention.

### Verification
- Focused `MediaCoverServiceFixture`: **32/32 passed**.
- Race-sensitive contention/deletion set: **30 repeated runs passed**.
- Complete Core suite: **4,441 passed, 26 skipped, 0 failed**.
- GitHub CI: **37 successful, 1 skipped, 0 failed**; only the repository's manual `Approve fork analysis` gate remains waiting.
- Exact two-file diff passed `git diff --check` and credential scanning with zero findings.
- An independent concurrency review found no remaining blocker after its follow-up recommendations were applied.

The PR remains open and unmerged for maintainer review.

plz12345 (approved):

PR #647, opened 2026-09-06
Fix: OpenAPI security requirements now reference their scheme definitions

alextomas955:
#### Database Migration

NO

#### Description

The served OpenAPI document declares its root `security` block as two empty objects, `[{}, {}]`,
instead of naming the two API-key schemes. An empty security requirement means "no credentials
required", so every consumer of the document concludes the API is unauthenticated: `/docs` renders
no padlocks and its Authorize dialog has no effect on requests, and generated clients
(openapi-generator, NSwag, Kiota) emit no auth parameter at all. The API itself still enforces the
key correctly; only what the document advertises is wrong.

**This is an Eros-only regression, and upstream already has the correct code.**

It was introduced here by 9371179fd, "Chore: Bump Swagger to 10.1.0" (#90). That commit moved to
the `AddSecurityRequirement(Func<OpenApiDocument, ...>)` overload and added the `doc` lambda
parameter, but never passed it to the reference constructor:

```diff
-                c.AddSecurityRequirement(new OpenApiSecurityRequirement
+                c.AddSecurityRequirement(doc => new OpenApiSecurityRequirement
                 {
-                    { apiKeyHeader, Array.Empty<string>() }
+                    { new OpenApiSecuritySchemeReference("X-Api-Key"), new List<string>() }
                 });
```

In Microsoft.OpenApi 2.7.5, pinned transitively by Swashbuckle.AspNetCore 10.2.3,
`OpenApiSecurityRequirement.SerializeInternal` filters each entry through
`CanSerializeSecurityScheme`, which keeps an entry only when the reference has a resolved `Target`
or when `Reference.HostDocument.Components.SecuritySchemes` contains the id. The single-argument
constructor leaves `HostDocument` null, so both entries are dropped and each requirement
serializes as `{}`.

`Whisparr/Whisparr` made the equivalent migration in c45758ec8 ("New: Bump .NET to 10", Whisparr/Whisparr#1197) and
passed the document through. This PR adopts that code verbatim rather than inventing a different
fix, so the block becomes textually identical to upstream and a future sync of this area is a
no-op instead of a conflict:

```diff
-                c.AddSecurityRequirement(doc => new OpenApiSecurityRequirement
+                c.AddSecurityRequirement(document => new OpenApiSecurityRequirement
                 {
-                    { new OpenApiSecuritySchemeReference("X-Api-Key"), new List<string>() }
+                    [new OpenApiSecuritySchemeReference(apiKeyHeader.Name, document)] = new List<string>(),
                 });
```

Before and after, read back from a running instance rather than transcribed:

| | Root `security` |
|---|---|
| Before | `[{}, {}]` |
| After | `[{"X-Api-Key": []}, {"apikey": []}]` |

Two separate requirement objects, not one object holding both keys. Two entries mean either scheme
is sufficient; one entry with two keys would mean both are required at once, which is not the
intent. The added test asserts that distinction rather than just checking both names appear.

Verified on a build from this branch:

- `GET /api/v3/system/status` returns 401 with no key, 200 with the key as an `X-Api-Key` header,
  and 200 with it as an `apikey` query parameter. The 401 is the control; without it the two 200s
  would not show the key was what let the calls through.
- On `/docs`, the Authorize dialog now lists both schemes by name and location. Entering a key
  flips the entry to Authorized and locks the operation padlocks, and "Try it out" on
  `GET /api/v3/system/status` returns 200. Before this change the same dialog had no effect and the
  call returned 401.
- `dotnet build src/NzbDrone.Host/Whisparr.Host.csproj -c Debug -p:Platform=Windows` reports 0
  warnings and 0 errors under the solution-wide `TreatWarningsAsErrors`.

One note on the build check. `./build.sh --backend` does not currently complete on this machine: it
fails at restore in about 1.3s with `NU1510` in `src/NzbDrone/Whisparr.csproj`, the WinForms tray
project. That failure is pre-existing, reproduces on a clean `eros-develop` with no changes, and
looks like an SDK feature-band mismatch, `global.json` pinning `10.0.101` against an installed
`10.0.400`. It is unrelated to this defect and that project has no code path to
`/docs/v3/openapi.json`, so I have left it alone rather than put a second unrelated change in this
PR. Happy to open a separate issue for it if that is useful.

<details>

Not a UI change. The `/docs` Authorize behaviour described above is a consequence of the corrected
document, not of any markup or component change.

</details>

#### Todos

- [x] Tests

`src/NzbDrone.Integration.Test/OpenApiSecurityFixture.cs` requests `/docs/v3/openapi.json` from a
running instance and asserts the served `security` array has exactly two entries, the first keyed
`X-Api-Key` and the second `apikey`. It follows `IndexHtmlFixture`'s pattern: a raw `HttpClient`
against `RootUrl` rather than the `api/v3/`-scoped REST client.

The test was checked for the failure it is supposed to catch. With the fix reverted and
`Whisparr.Console` rebuilt, it fails with `Expected collection to contain a single item matching
(property.Name == "X-Api-Key"), but the collection is empty`. With the fix in place it passes.

No translation keys are needed: this change adds no user-facing strings.

- Fixes whisparr/whisparr#1252

### DIFF
--- a/src/NzbDrone.Host/Startup.cs
+++ b/src/NzbDrone.Host/Startup.cs
@@ -131,7 +131,7 @@ namespace NzbDrone.Host
 
                 c.AddSecurityRequirement(doc => new OpenApiSecurityRequirement
                 {
-                    { new OpenApiSecuritySchemeReference("X-Api-Key"), new List<string>() }
+                    { new OpenApiSecuritySchemeReference("X-Api-Key", doc), new List<string>() }
                 });
 
                 var apikeyQuery = new OpenApiSecurityScheme
@@ -157,7 +157,7 @@ namespace NzbDrone.Host
 
                 c.AddSecurityRequirement(doc => new OpenApiSecurityRequirement
                 {
-                    { new OpenApiSecuritySchemeReference("apikey"), new List<string>() }
+                    { new OpenApiSecuritySchemeReference("apikey", doc), new List<string>() }
                 });
 
                 c.DescribeAllParametersInCamelCase();
--- a/src/NzbDrone.Host/Startup.cs
+++ b/src/NzbDrone.Host/Startup.cs
@@ -129,9 +129,9 @@ namespace NzbDrone.Host
 
                 c.AddSecurityDefinition("X-Api-Key", apiKeyHeader);
 
-                c.AddSecurityRequirement(doc => new OpenApiSecurityRequirement
+                c.AddSecurityRequirement(document => new OpenApiSecurityRequirement
                 {
-                    { new OpenApiSecuritySchemeReference("X-Api-Key", doc), new List<string>() }
+                    [new OpenApiSecuritySchemeReference(apiKeyHeader.Name, document)] = new List<string>(),
                 });
 
                 var apikeyQuery = new OpenApiSecurityScheme
@@ -155,9 +155,9 @@ namespace NzbDrone.Host
 
                 c.AddSecurityDefinition("apikey", apikeyQuery);
 
-                c.AddSecurityRequirement(doc => new OpenApiSecurityRequirement
+                c.AddSecurityRequirement(document => new OpenApiSecurityRequirement
                 {
-                    { new OpenApiSecuritySchemeReference("apikey", doc), new List<string>() }
+                    [new OpenApiSecuritySchemeReference(apikeyQuery.Name, document)] = new List<string>(),
                 });
 
                 c.DescribeAllParametersInCamelCase();
--- a/src/NzbDrone.Integration.Test/OpenApiSecurityFixture.cs
+++ b/src/NzbDrone.Integration.Test/OpenApiSecurityFixture.cs
@@ -1,0 +1,27 @@
+using System.Net.Http;
+using System.Text.Json;
+using FluentAssertions;
+using NUnit.Framework;
+
+namespace NzbDrone.Integration.Test
+{
+    [TestFixture]
+    public class OpenApiSecurityFixture : IntegrationTest
+    {
+        private HttpClient _httpClient = new HttpClient();
+
+        [Test]
+        public void should_serve_both_api_key_schemes_in_root_security()
+        {
+            var request = new HttpRequestMessage(HttpMethod.Get, RootUrl + "docs/v3/openapi.json");
+            var response = _httpClient.Send(request);
+            var text = response.Content.ReadAsStringAsync().GetAwaiter().GetResult();
+
+            var security = JsonDocument.Parse(text).RootElement.GetProperty("security");
+
+            security.GetArrayLength().Should().Be(2);
+            security[0].EnumerateObject().Should().ContainSingle(property => property.Name == "X-Api-Key");
+            security[1].EnumerateObject().Should().ContainSingle(property => property.Name == "apikey");
+        }
+    }
+}
--- a/src/NzbDrone.Integration.Test/OpenApiSecurityFixture.cs
+++ b/src/NzbDrone.Integration.Test/OpenApiSecurityFixture.cs
@@ -1,3 +1,6 @@
+using System.Collections.Generic;
+using System.Linq;
+using System.Net;
 using System.Net.Http;
 using System.Text.Json;
 using FluentAssertions;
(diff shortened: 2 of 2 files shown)


plz12345 (changes requested):
Thanks for this — the diagnosis is right, the fix is minimal, and adopting upstream's exact form so a future sync is a no-op is the correct call. I verified it end-to-end locally rather than taking the description on trust:

- Pre-fix build serves `security: [{}, {}]`; a build of this branch serves `security: [{"X-Api-Key": []}, {"apikey": []}]`. Two separate requirement objects, as intended.
- `GET /api/v3/system/status` on the branch build: 401 with no key, 200 with the `X-Api-Key` header, 200 with the `apikey` query parameter. The 401 is the control — auth behaviour is unchanged.
- The test is not vacuous. With the fix reverted and `Whisparr.Console` rebuilt it fails with `Expected collection to contain a single item matching (property.Name == "X-Api-Key"), but the collection is empty`; restored, it passes.
- `Whisparr.Console` builds with 0 warnings, 0 errors under solution-wide `TreatWarningsAsErrors`.

Requesting changes on four small things, none of which touch the fix itself.

---

**1. `OpenApiSecurityFixture.cs:19` — assert the status code before parsing.**

If `/docs/v3/openapi.json` ever 404s or 500s, this fails as a `JsonException` on the response body rather than saying what actually went wrong. One line up front:

```csharp
response.StatusCode.Should().Be(HttpStatusCode.OK);
```

**2. `OpenApiSecurityFixture.cs:23-24` — drop the positional coupling.**

`security[0]` / `security[1]` bind the test to the order of the two `AddSecurityDefinition` calls in `Startup.cs`. Reordering those is behaviour-neutral but would fail this test, which makes it a tripwire for something it isn't trying to protect.

The distinction you're actually asserting — two separate requirements, not one object holding both keys — is already carried by `GetArrayLength().Should().Be(2)`. Please make the identity check order-independent, e.g. collect the single property name from each entry and assert the set contains both `X-Api-Key` and `apikey`.

**3. `OpenApiSecurityFixture.cs:12` — `_httpClient` should be `readonly`, and dispose the response.**

I know this mirrors `IndexHtmlFixture`, and that was a reasonable file to model on for the raw-`HttpClient`-against-`RootUrl` approach. But this is a new file, and new code is held to current house style regardless of what it was copied from — the pre-existing instance is grandfathered, a fresh one isn't. `sonar.test.exclusions` in `sonarqube.yml` covers binaries only, so test sources are analysed and this will be flagged as new code.

```csharp
private readonly HttpClient _httpClient = new HttpClient();
```

and `using var response = _httpClient.Send(request);`.

**4. Please drop the `🤖 Generated with Claude Code` footer from the PR description.**

It isn't part of `.github/PULL_REQUEST_TEMPLATE.md`, and PRs are the project's record of work rather than advertising space. To be clear about scope: your commits are clean — all three are authored and committed by you with no `Co-Authored-By` trailer — and this repo squashes with `squash_merge_commit_message: BLANK`, so the footer can't reach the merged commit or contributor stats either way. This is about the PR page itself.

---

Two non-blocking notes.

**SonarQube new-code coverage may score 0% on this change, and that isn't your fault.** Coverage is downloaded from `build_v3.yml`'s Windows `unit_test` leg, which filters on `TestCategory!=IntegrationTest`. The only test covering the four changed lines in `Startup.cs` is an integration test, so nothing offsets them in the report. An integration test is the right instrument here — `ConfigureServices` isn't meaningfully unit-testable — so don't chase this. If the gate fires we'll waive it on the Sonar side.

**A reproduction caveat for anyone re-checking the revert.** `dotnet test` on `Whisparr.Integration.Test` does not rebuild `Whisparr.Host` — the test project doesn't reference it, and the runner launches whatever is already in `_output`. My first revert run passed against a stale binary before I caught it. An explicit `dotnet build src/NzbDrone.Console/Whisparr.Console.csproj` is needed between the edit and the test. CI is unaffected; it builds the solution first.

CI hasn't run yet — you're a first-time contributor, so `build_v3` and SonarQube are sitting behind GitHub's fork-approval gate. That's a workflow step, not a problem with the PR; I'll approve the runs alongside the next push.

alextomas955:
Thanks for verifying it end-to-end rather than taking the description on trust, and for catching
the stale-binary trap yourself — that one cost me a wrong RED run earlier too.

All four addressed in e93d55004.

**1. Status code before parsing.** Added `response.StatusCode.Should().Be(HttpStatusCode.OK);`
ahead of reading the body.

**2. Positional coupling dropped.** Each requirement is still asserted to name exactly one scheme,
so the two-alternatives-versus-both-required distinction survives, but the names are now matched as
an unordered set:

```csharp
requirements.Should().HaveCount(2);
requirements.Should().OnlyContain(schemeNames => schemeNames.Count == 1);

requirements.SelectMany(schemeNames => schemeNames)
    .Should().BeEquivalentTo(new List<string> { "X-Api-Key", "apikey" });
```

Reordering the two `AddSecurityDefinition` calls now passes, which is the behaviour you asked for.

**3. `readonly` and response disposal.** Both done. Fair point on new code not inheriting the
grandfathered pattern from `IndexHtmlFixture` — I had modelled on it for the raw-`HttpClient`
approach and carried the field declaration along with it, which I should not have.

**4. Footer removed** from the PR description.

Since the assertions changed, the earlier red/green evidence no longer applied, so I re-ran it with
the explicit `Whisparr.Console` rebuild between steps:

- Fix reverted: `Failed` — `Expected collection to contain only items matching
  (schemeNames.Count == 1), but {{empty}, {empty}} do(es) not match.`
- Fix restored: `Passed! - Failed: 0, Passed: 1`

The new message is actually more diagnostic than the old one, since it prints the two empty
requirement objects that are the defect.

`Whisparr.Integration.Test` builds 0 warnings / 0 errors under solution-wide
`TreatWarningsAsErrors`.

Noted on both non-blocking points, and thanks for pre-empting the Sonar coverage question — I will
leave that alone rather than adding a unit test that does not really exercise `ConfigureServices`.

PR #882, opened 2026-09-17
chore: Declare the response type on twelve API actions

alextomas955:
#### Database Migration

NO

#### Description

Twelve actions return `object`, so Swashbuckle has nothing to infer from and the served
document describes each one as `200` with no content. A generated client gets a method that
returns nothing, for endpoints that have always returned data. `GET /api/v3/lookup/scene` is
the clearest example.

Each action now declares the type it already returns:

```diff
-public object SearchScene([FromQuery] string term)
+public List<SearchResource> SearchScene([FromQuery] string term)
```

Only the static type changes, so `ObjectResultExecutor` resolves the same object type and the
JSON on the wire is unchanged.

#### What was captured, and what was not

Nine of the twelve were captured from a running instance built from this branch and from one
built from `eros-develop`, both starting from an empty data directory, and compared byte for
byte. All nine identical:

| | |
|---|---|
| `GET /api/v3/autotagging/schema` | 13,067 bytes |
| `GET /api/v3/customformat/schema` | 23,558 bytes |
| `GET /api/v3/config/naming/examples` | 516 bytes |
| `GET /api/v3/importlist/movie` | 2 bytes |
| `GET /api/v3/lookup/scene` | 38,101 bytes |
| `GET /api/v3/lookup/movie` | 144,807 bytes |
| `GET /api/v3/lookup/studio` | 15,246 bytes |
| `GET /api/v3/lookup/performer` | 13,765 bytes |
| `POST /api/v3/exclusions/bulk` | 145 bytes |

`exclusions/bulk` is a POST that reaches the changed return and comes back with the mapped
resource, so a mutating path is covered rather than only reads.

The remaining two were **not** captured. `POST /api/v3/release` needs a populated remote movie
cache and `POST /api/v3/manualimport` needs an existing movie, so neither reaches its return on
an offline instance.

They do not need a capture, and the reason is stronger than one. Each returns its own
`[FromBody]` parameter:

```csharp
public async Task<ReleaseResource> DownloadRelease([FromBody] ReleaseResource release)
    ... return release;

public List<ManualImportReprocessResource> ReprocessItems([FromBody] List<ManualImportReprocessResource> items)
    ... return items;
```

The value returned is the model-bound instance of the declared type. Model binding constructs
exactly that type, so no subtype is reachable and declared and runtime types cannot diverge.

An earlier version of this description and the commit message both listed endpoints that had
not actually been captured. Both now say the same thing, and it is the list above.

#### Precedent

Sonarr already declares `IEnumerable<SeriesResource>` on the equivalent lookup action in its
**live V3 API**, in `Sonarr.Api.V3/Series/SeriesLookupController.cs`. These four `lookup/*`
actions are catching up with that, not going beyond it.

My earlier citation of Sonarr v5 was loose and I have dropped it. Their V3 controllers still
return `object` exactly as ours did, and the v5 typing is `Ok<List<...>>`, a Minimal API
`HttpResults` idiom from their rewrite that is not a shape to copy into V3.

Actions touched: `AutoTaggingController.GetTemplates`, `CustomFormatController.GetTemplates`,
`NamingConfigController.GetExamples`, `ImportListExclusionController.AddImportListExclusions`,
`ImportListMoviesController.GetDiscoverMovies` and `.AddMovies`,
`ReleaseController.DownloadRelease`, `ManualImportController.ReprocessItems`, and the four
`SearchController` lookups.

<details>

Not a UI change.

</details>

#### Todos

- [x] Tests

No new tests. The declared type is the guarantee: a method can no longer drift from what it
returns without the compiler objecting. `Whisparr.Api.Test` passes 142/142.

No issue filed. This came out of a Discord conversation about making the OpenAPI document
accurate enough to generate a client from.

Related to #887


plz12345 (commented):
Reviewed the diff and read each touched method in full. All twelve declare the type they already returned on every non-throwing path, and the runtime type is unchanged, so `ObjectResultExecutor` resolves the same `objectType` it did before — the wire format claim holds. CI is green across the backend matrix, unit, postgres and integration suites.

Two things to tidy before merge.

**1. The two mutating endpoints were never captured.** The PR body lists `autotagging/schema`, `customformat/schema`, `config/naming/examples`, `exclusions/bulk`, `importlist/movie`, `exclusions`, `lookup/scene`; the commit message gives a different list. Neither includes `POST /api/v3/release` or `POST /api/v3/manualimport`. I don't think either is at risk — `DownloadRelease` returns the deserialized `release` argument and `ReprocessItems` returns the mutated `items` list, so declared and runtime types match — but the PR's safety argument is "I checked that rather than assumed it," and those two weren't checked. Either capture them or say plainly they were reasoned about. Worth reconciling the commit body with the PR body too, since the commit message is what stays in the log.

**2. The Sonarr v5 precedent is looser than it reads.** Sonarr's V3 controllers still return `object`, identical to ours; the typing exists only in `Sonarr.Api.V5`, and there it's `Ok<List<...>>` rather than a bare `List<...>`. The inner types match, the declared types don't. `Ok<T>` is a Minimal API `HttpResults` idiom from their rewrite and isn't a shape we want in V3, so the choice here is right — the citation is just imprecise, and I'd rather it not read as "follow V5's shape."

There's a stronger precedent available for the lookup endpoints: `Sonarr.Api.V3/Series/SeriesLookupController.cs` already declares `IEnumerable<SeriesResource> Search(...)` in their live V3 API. The four `lookup/*` actions here aren't ahead of Sonarr, they're catching up.

No tests requested. The declared type *is* the guarantee — a method can no longer drift from what it returns without the compiler objecting, which is stronger than a shape assertion, and `integration_test` already exercises these paths.

Note for whoever merges: SonarQube hasn't run (`Approve fork analysis` is pending on the fork gate), so that gate is unmeasured here.

alextomas955:
Both fair. Force-pushed 038be6f53 with the commit message rewritten and the description to match.

**1. The mutating endpoints.** You were right that they were not checked, and checking properly
turned up worse: the old commit message and description each listed endpoints that had never been
captured at all. The commit claimed `lookup/movie`, `lookup/studio` and `lookup/performer`; the
description claimed `exclusions/bulk`. Six were captured, not nine. That was sloppy and I have
corrected it rather than papering over it.

Nine are now genuinely captured, from this branch and from `eros-develop`, both starting from an
empty data directory so ids line up, compared byte for byte and all identical:

```
autotagging/schema     13,067    lookup/scene       38,101
customformat/schema    23,558    lookup/movie      144,807
config/naming/examples    516    lookup/studio      15,246
importlist/movie            2    lookup/performer   13,765
POST exclusions/bulk      145
```

`exclusions/bulk` is a POST that reaches the changed return and comes back with the mapped
resource, so that covers a mutating path rather than only reads.

`POST /api/v3/release` and `POST /api/v3/manualimport` I could not drive offline. One needs a
populated remote movie cache, the other an existing movie; both throw before reaching the return.
Rather than assert they are fine, the description now gives the reason they cannot vary: each
returns its own `[FromBody]` parameter, `release` and `items`. The returned value is the
model-bound instance of the declared type, so no subtype is reachable and declared and runtime
types cannot diverge. That is a compile-time property, not a reasoned guess, and I would rather
state it than claim a check I did not run.

**2. The Sonarr citation.** Also right, and I have dropped the v5 reference entirely. Their V3
controllers return `object` exactly as ours did, and `Ok<List<...>>` is a Minimal API
`HttpResults` idiom from the rewrite, not a shape to bring into V3.

`Sonarr.Api.V3/Series/SeriesLookupController.cs` declaring `IEnumerable<SeriesResource> Search(...)`
is the better precedent and the description now uses it. It reframes the four `lookup/*` changes
as catching up with their live V3 rather than getting ahead of anything, which is the honest
framing.

Noted on tests and on the fork gate leaving SonarQube unmeasured.

alextomas955:
Your closing note is now resolved: the force-push cleared the fork gate, so `Approve fork analysis`
is green and SonarQube did run. Quality gate passed, 0 new issues, 0 accepted issues, 0 security
hotspots. The only non-success check left is `deploy / release`, which is skipped by design on a PR.

So that gate is measured rather than unmeasured for whoever picks this up.

plz12345 (approved):
Approved. One accounting note, not blocking: nine captured plus two reasoned-about makes eleven. `ImportListMoviesController.AddMovies` (`POST /api/v3/importlist/movie`) is in neither list. The compiler already pins it to `List<MovieResource>`, so it's covered anyway.

PR #883, opened 2026-09-17
chore: Name the response types returned as anonymous objects

alextomas955:
#### Database Migration

NO

#### Description

Seven actions return an anonymous type. An anonymous type cannot be described, so the served
document declares each one as `200` with no content.

Each now returns a named resource. The names and shapes are taken from Sonarr's v5 API, which
has already done this work, rather than invented here:

| Whisparr | Sonarr v5 |
|---|---|
| `SystemShutdownResource` | same |
| `SystemRestartResource` | same |
| `BackupRestoreResource` | same |
| `FileSystemTypeResource` | same |
| `FileSystemMediaFileResource` | same |
| `MovieFolderResource` | `SeriesFolderResource` |

The names match v5 exactly. Nothing upstream reaches these files, `Sonarr.Api.V5` commits are
skipped here and Sonarr's V3 `FileSystemController` still returns `new { type = "file" }`, so
this is consistency with v5 rather than avoiding a future conflict.

`FileSystemTypeResource.Type` carries the existing `FileSystemEntityType` enum rather than a
loose string, again matching upstream. It serialises to the same `"file"` and `"folder"`.

`GetMediaFiles` is included because it was inconsistent rather than merely undescribed: it
returned `string[]` when the folder was missing and an array of anonymous objects otherwise,
so one endpoint had two shapes. It now returns `FileSystemMediaFileResource` either way, empty
for a missing folder, which is what upstream settled on.

The wire shape is unchanged throughout, checked against a running instance:

```
filesystem/type       {"type":"file"} and {"type":"folder"}
filesystem/mediafiles [] for a missing folder, otherwise the same
                      path / relativePath / name objects
```

The resources deliberately do not derive from `RestResource`. That carries an `Id`, which
would have added an `id` field that was never in these responses. Plain resource classes are
already common here, 25 of them against 49 deriving `RestResource`.

Seven operations stop reporting an empty `200`. Measured against `eros-develop` at
6280c42d6 the document-wide count goes from 81 to 74.

<details>

Not a UI change.

</details>

#### Todos

- [x] Tests

No new tests. The wire format is unchanged and the check is the live capture above.
`Whisparr.Api.Test` passes 142/142.


plz12345 (commented):
Wire format holds. I checked the two value-level changes: `FileSystemEntityType.File` serialises as `"file"` because `STJson` registers `JsonStringEnumConverter(JsonNamingPolicy.CamelCase)`, so `PathInput.tsx` and `FileBrowserRow.tsx` keep matching; and both empty-folder returns in `GetMediaFiles` serialise to `[]`. The other five are same-named property swaps. Skipping `RestResource` to avoid a phantom `id` is right.

One correction to the description: matching Sonarr v5's names does not keep "a future sync of this area a no-op", because no sync reaches it. `Sonarr.Api.V5` commits are skipped here (we have no V5 project), and Sonarr's V3 `FileSystemController` still returns `new { type = "file" }`. Nothing upstream will ever land in these files. The names do match v5 exactly, and consistency with it is a fine reason by itself, so it's just the sync claim that should go. Same point as the v5 citation on #882.

Nits, not blocking:
- `FileSystemEntityType` has four members, so the schema will advertise `parent` and `drive` for an endpoint that only ever returns `file` or `folder`. Acceptable as a deliberate match with upstream.
- `FileSystemEntityType? Type` is nullable but always set, so the schema says `nullable: true`. The `?` is carried over from v5, where nullable reference types are on.

plz12345:
Thanks for these. Rather than review seven PRs separately, I've pulled all of them into #893 with your authorship kept, plus a few small follow-ups on top. Closing this in favor of that one. Future typing work can go against #887.

PR #885, opened 2026-09-17
chore: Mark the anonymous endpoints as taking no credentials

alextomas955:
#### Database Migration

NO

#### Description

The document declares a root security requirement and nothing overrides it, so every operation
claims to need an API key. Six of them do not:

```
GET /ping          GET /login       GET /logout
HEAD /ping         POST /login      GET /content/{path}
```

All six are `[AllowAnonymous]` and answer without a key. A generated client sends credentials
it does not need, and anyone reading the document is told these are authenticated when they
are not.

An operation filter clears the requirement where the action carries `IAllowAnonymous`. An empty
list is how OpenAPI spells "this operation takes no credentials".

The filter reads endpoint metadata rather than keeping a list of paths, so an action that gains
or loses `[AllowAnonymous]` is picked up without anyone remembering to edit this.

Checked against a running instance with no key at all:

```
/ping            200      /api/v3/movie   401
/login           200      /api            401
/logout          302
/content/...     reachable
```

`GET /api` looked like a seventh case but is not. It has no `[AllowAnonymous]`, answers 401
without a key, and is correctly left inheriting the root requirement.

This is the first `IOperationFilter` in the codebase. It is the Swashbuckle extension point for
per-operation document changes, and there is no attribute-based way to express this, so there
was no lighter option.

Radarr and Sonarr do not mark these, including Sonarr's v5 API. Whisparr V2 has since added the
same filter in Whisparr/Whisparr#1310, so the class name and folder here match it. Nothing syncs
between the two repos, so that is consistency across the forks rather than avoiding a conflict.

<details>

Not a UI change.

</details>

#### Todos

- [x] Tests

No new tests. The change affects only the generated document, and the check is the live
comparison above. `Whisparr.Api.Test` passes 142/142.


plz12345:
Heads up on merge order: this and #886 both branch from the same `Startup.cs` blob (`62f1b2b3fc`) and insert a different line at the identical anchor — after the security-requirement block, immediately before `c.DescribeAllParametersInCamelCase()`:

```
#885:  + c.OperationFilter<AnonymousOperationFilter>();
#886:  + c.SchemaFilter<CommandArgumentsSchemaFilter>();
```

Git can't auto-merge two different insertions at the same point, so whichever lands second will conflict. Nothing wrong with either PR — just expect a rebase on the second one. Eros PRs are squash-merged, so that rebase is `git rebase --onto <squash-commit> <old-base>`, not a plain rebase against `eros-develop`.

No action needed now; flagging so it isn't a surprise.

PR #886, opened 2026-09-17
chore: Let the command schema carry a command's arguments

alextomas955:
#### Database Migration

NO

#### Description

`CommandController.StartCommand` reads the request body a second time and deserialises it into
the concrete command type:

```csharp
Request.Body.Seek(0, SeekOrigin.Begin);
var body = reader.ReadToEnd();
var command = STJson.Deserialize(body, commandType) as Command;
```

So a command's own arguments travel at the top level, alongside the `CommandResource` fields.
Swashbuckle closes every object schema with `additionalProperties: false`, which forbids
exactly those. The document therefore tells a client it may not send arguments to any of the 42
commands, when in fact that is the only way to send them. The `body` property looks like the
place for them, but nothing reads it.

A schema filter reopens `CommandResource`. Comparing the served document before and after: one
schema changed, none added or removed.

Checked against a running instance:

```
POST /api/v3/command
{"name":"RefreshMovie","movieIds":[],"isNewMovie":true}

-> body.isNewMovie = true
```

`isNewMovie` defaults to `false`, so the value came off the top level rather than being
ignored. No server change, the payload was always accepted.

#### What this does not do

It says arguments are allowed. It does not say which arguments each command takes. Doing that
means a `oneOf` across the 42 command types with a discriminator on `name`, which is a much
larger change and worth agreeing on separately.

Worth knowing: Sonarr's v5 API has the same defect. Its `CommandController` still re-reads the
body and its committed `openapi.json` still carries `additionalProperties: false` on
`CommandResource`. Whisparr V2 has since fixed it the same way in Whisparr/Whisparr#1310, so
the class name, folder and filter body here match it. Nothing syncs between the two repos, so
that is consistency across the forks rather than avoiding a conflict.

This is an `ISchemaFilter`, the Swashbuckle extension point for schema changes, and the first in
the codebase. `additionalProperties: false` is Swashbuckle's default for every object schema,
so there is no way to exempt one type without one.

<details>

Not a UI change.

</details>

#### Todos

- [x] Tests

No new tests. The change affects only the generated document, and the check is the live request
above. `Whisparr.Api.Test` passes 142/142.


plz12345:
Heads up on merge order: this and #885 both branch from the same `Startup.cs` blob (`62f1b2b3fc`) and insert a different line at the identical anchor — after the security-requirement block, immediately before `c.DescribeAllParametersInCamelCase()`:

```
#885:  + c.OperationFilter<AnonymousOperationFilter>();
#886:  + c.SchemaFilter<CommandArgumentsSchemaFilter>();
```

Git can't auto-merge two different insertions at the same point, so whichever lands second will conflict. Nothing wrong with either PR — just expect a rebase on the second one. Eros PRs are squash-merged, so that rebase is `git rebase --onto <squash-commit> <old-base>`, not a plain rebase against `eros-develop`.

No action needed now; flagging so it isn't a surprise.

PR #902, opened 2026-09-19
New: Parse PRT release group from scene titles without a dash

### LINKED ISSUE #676: Parser: recognise release groups PRT (".PRT") and GhostFreakXX ("[GhostFreakXX]") that don't use the "-GROUP" convention
### What branch are you running?

Eros (V3 "purple")

### Is there an existing issue for this?

- [x] I have searched the existing open and closed issues

### Is your feature request related to a problem? Please describe

Whisparr's release group parsing expects the group at the end of a release name in the "-GROUP" format. Two of the most active HEVC release groups use different conventions:

- PRT appends the group with a dot: `.PRT`
- GhostFreakXX appends the group in square brackets: `[GhostFreakXX]`

Because of this, the parser returns no release group for their releases. For example, the parse API (`/api/v3/parse?title=...`) returns `releaseGroup: null` for:

    Studio.26.01.01.Model.Name.Some.Random.Scene.XXX.1080p.HEVC.x265.PRT

and likewise for GhostFreakXX releases such as:

    [Studio] (Model Name) Some Random Scene XXX (2026) (1080p HEVC) [GhostFreakXX]

This causes several problems:

1. On import, the release group is stored as empty, so the `{-Release Group}` naming token renders nothing and the group is lost from the renamed file. With the naming format `{Studio CleanTitleSlug} {Release-Date} {Scene CleanTitle} [{Quality Title}]{-Release Group}`.
2. Custom formats using a "Release Group" condition never match these releases. The only workaround is a "Release Title" regex, which is less precise.
3. Users relying on the release group for upgrades, scoring or library organisation get inconsistent results for the two main HEVC groups.

### Describe the solution you'd like

Add PRT and GhostFreakXX to the parser's special-case handling for release groups that don't follow the "-GROUP" convention (similar to the existing exceptions inherited from Radarr), so that:

- A trailing `.PRT` at the end of a release name is parsed as release group `PRT`
- A trailing `[GhostFreakXX]` at the end of a release name is parsed as release group `GhostFreakXX`

Matching should be anchored to the end of the title (before any file extension) so these strings aren't picked up if they appear elsewhere in a scene title.

Since both groups are currently the main source of HEVC releases, this would fix automatic imports, renaming and Release Group custom format conditions for a large share of releases without any user workaround.

### Describe alternatives you've considered

- **Manual Import:** Setting the Release Group field manually in the Manual Import dialog works, but isn't practical for automated downloads.
- **Renaming in qBittorrent before import:** I tried a script triggered by qBittorrent's "Run external program on torrent finished" that uses the qBittorrent Web API to rename the torrent's file and folder (e.g. `.PRT` → `-PRT`) so the torrent keeps seeding and Whisparr can still match it by hash. However, this creates a race condition: Whisparr's Completed Download Handling often imports the file before the script finishes renaming, so the original name is parsed anyway and the group is still lost.
- **Custom formats with Release Title regex:** Matching `\bPRT\b` / `\[GhostFreakXX\]` against the release title allows scoring to work, but doesn't populate the release group, so renaming still drops it.

None of these fix the underlying parsing, which is why built-in handling in the parser seems like the right solution.

### Anything else?

_No response_

plz12345:
#### Database Migration

NO

#### Description

PRT ends its releases with `.PRT` (usually `.PRT[XvX]`) rather than `-PRT`, so the release group came back empty — nothing for `{Release Group}` in renames or Release Group custom format conditions. Adds `PRT` to the exact-match exception list, anchored to the end of the title (an optional trailing `[tag]` allowed), so a mid-title "PRT" doesn't override a real `-GROUP`.

The GhostFreakXX half of the issue already parses; covered by a test to keep it that way.

Verified: all 1,334 ParserTests pass; the new PRT cases fail without the change.

<details>

</details>

#### Todos

- [x] Tests

- Fixes #676


bradenmitchell:
@plz12345 Appreciate your work, thanks.

Issue #887, opened 2026-09-17
OpenAPI: type the API actions still returning object


plz12345:
#882, #883 and #884 typed most of the API actions that previously declared `object` or returned an anonymous object, so the served OpenAPI document describes them properly. They don't finish the surface. This tracks the remainder so it doesn't stall partway.

### Still untyped

**`ProviderControllerBase.cs:181,190,223`** — `DeleteProvider`, `DeleteProviders`, `Test`. This is the important one: the class is generic, so these are the declared shape for *every* indexer, download client, notification, metadata and import-list provider endpoint. Leaving it means a large share of the document stays empty no matter what the merged PRs did.

- `MetadataController.cs:28` and `NotificationController.cs:27` — `DeleteProviders` overrides, inheriting the above.
- `CustomFormatController.cs:110` — `DeleteFormats`. #882 typed `GetTemplates` in this same file and left its neighbour.
- `BlocklistController.cs:74` — `Remove`
- `ImportListExclusionController.cs:148` — `DeleteImportListExclusions`
- `HistoryController.cs:153` — `MarkAsFailed`
- `QueueController.cs:98` — `RemoveMany`
- `SystemController.cs:123` — `DuplicateRoutes`
- `BackupController.cs:53` — `DeleteBackup`

### Anonymous `{}` bodies

These return `new { }` — a 200 with an empty object — and fall in the gap between #883's "name the anonymous objects" and #884's "document the 202s":

- `MovieFileController.cs:204` — `DeleteMovieFiles`
- `MovieEditorController.cs:125` — `DeleteMovies`
- `PerformerEditorController.cs:121` — `DeletePerformers`

### Note on upstream

This is not a sync candidate. Sonarr did the equivalent work inside `Sonarr.Api.V5`, which `UPSTREAM_SYNC.md` correctly dispositions as skip — V5 files land dead here. Their V3 controllers still return `object` for the same actions. So this has to be done by hand against our live V3 surface, and `Ok<T>` (their Minimal API `HttpResults` idiom) is not the shape to copy; bare `List<T>` / `ActionResult<T>` is correct for our MVC controllers.

One exception worth knowing: `Sonarr.Api.V3/Series/SeriesLookupController.cs` has always declared `IEnumerable<SeriesResource>`, so on lookup endpoints we were behind their live API rather than ahead of it.

PR #905, opened 2026-09-19
chore: Finish the OpenAPI accuracy pass

### LINKED ISSUE #904: OpenAPI: the served document still misdescribes 50 operations and does not validate
#882 and #893 typed the actions that declared `object` or returned anonymous objects, and #898 and #900 finished that surface. What is left is a different set of inaccuracies: the document is wrong about status codes, media types, request bodies and paths, and as a whole it does not validate. This tracks the remainder.

Everything below is a fault in the generated document. No response on the wire is wrong.

### The document does not validate

Feeding `/docs/v3/openapi.json` to a validator fails outright:

```
Path parameter 'path' for 'get' operation in '/' was not resolved
```

`StaticResourceController.Index` carries two route attributes, `[HttpGet("")]` and `[HttpGet("/{**path:regex(...)}")]`. Swashbuckle emits one operation per route and both inherit the method's `[FromRoute] string path`, so the operation for the empty template declares a required path parameter its own template has no slot for.

A validator rejects the whole file on this, not just that operation, so any consumer that checks the document before using it is blocked on everything.

### Two paths are declared twice

```
/api/v3/performer/{id}   and  /api/v3/performer/{performerForeignId}
/api/v3/studio/{id}      and  /api/v3/studio/{studioForeignId}
```

The OpenAPI spec says templated paths with the same hierarchy but different templated names must not both exist, because they are the same path. ASP.NET separates them at runtime with the `{id:int}` constraint, which OpenAPI cannot express, so a reader of the document cannot tell which one `/performer/5` refers to. `MovieController` already hit this and solved it.

### Every REST-convention action is wrong about its status code

`RestController.Created` answers 201 and `RestController.Accepted` answers 202, but ApiExplorer only sees the declared return type and documents a 200 for both. That is **46 operations**: 18 that answer 201, 26 that answer 202, and two that already declare the right code alongside a stray 200.

`POST /api/v3/tag` is the simplest case to see: the document says 200, the server answers 201.

### Request bodies all claim to be optional

All 93 of them carry `required: false`. No `[FromBody]` parameter in the codebase is nullable or has a default, so a request omitting the body is rejected with a 400 before the action runs. ApiExplorer reports them optional only because nullable reference types are off in these projects.

```
POST /api/v3/tag  with no body  ->  400
```

### Five non-JSON endpoints are documented as returning nothing

| endpoint | actually returns |
|---|---|
| `GET /api/v3/system/routes` | `text/plain` |
| `GET /api/v3/log/file/{filename}` | `text/plain` |
| `GET /api/v3/log/file/update/{filename}` | `text/plain` |
| `GET /feed/v3/calendar/whisparr.ics` | `text/calendar` |
| `GET /api/v3/mediacover/{movieId}/{filename}` | an image |

A return type cannot carry a media type, so each needs `[Produces]` and a `[ProducesResponseType]`.

### Two provider endpoints answer with a body the document ignores

Both are on `ProviderControllerBase`, so each affects all five provider controllers.

`TestAll` builds a `List<ProviderTestAllResult>` and returns it as 200, or as 400 when any provider fails validation. The document declares an empty 200, no 400 at all, and `ProviderTestAllResult` does not appear in the document.

`RequestAction` returns `Content(data.ToJson(), "application/json")`. The shape varies per action, so a free-form schema is the honest answer, but the document currently says no body comes back.

### Redirects and the backup upload

`POST /login` and `GET /logout` answer 302 and are documented as 200, a status neither returns. Login also answers 401 with `ProblemDetails` when `SignInAsync` throws.

`POST /api/v3/system/backup/restore/upload` requires a multipart file and declares no request body at all, so there is no way to call it from the document. `UploadAndRestore` reads `Request.Form.Files` directly instead of binding a parameter, so ApiExplorer sees nothing to describe.

### Not covered here

**Error responses.** Across 274 operations, two declare a 400 and six declare a 404, while the API returns `application/problem+json` widely. Doing this properly means agreeing a shape and touching nearly every operation, so it belongs on its own.

**Command arguments.** #886 opened `CommandResource` so the document stops forbidding them, but no per-command shape exists. That needs a `oneOf` with a discriminator and has a design question attached, so it is filed separately.

**`POST /api/v3/system/backup/restore/upload` answers 500 on a missing body.** `Request.Form` throws before the action's own `BadRequestException` guard runs, so the intended 400 is unreachable. That is a behaviour fix rather than a documentation one.

Related to #882, #887, #893, #898 and #900.

alextomas955:
#### Database Migration

NO

#### Description

The remainder of the OpenAPI work from #882 and #893.

The served document did not validate. `StaticResourceController.Index` carries two route attributes, and the empty one has no `{path}` slot, so the operation generated for it declared a required path parameter that nothing could resolve:

```
Path parameter 'path' for 'get' operation in '/' was not resolved
```

- A validator rejects the whole file on that rather than the single operation, so anything checking the document before using it stopped there. The controller is excluded rather than that one action: all three of its actions serve the frontend through `MapResource`, the SPA shell, its static assets and the login page, so none of it is API surface. Excluding only `Index` would have left `/content/{path}` and `GET /login` documented while `/{path}` was not, which is an arbitrary line. The controller itself is otherwise untouched, and `ApiExplorerSettings` only feeds document generation.
- Two paths were also declared twice. `/api/v3/performer/{id}` and `/api/v3/performer/{performerForeignId}` are the same path as far as OpenAPI is concerned and the spec forbids that, and the same applies to studio. ASP.NET tells them apart at runtime with the `{id:int}` constraint, which OpenAPI has no way to express, so a reader of the document cannot know which one `/performer/5` means. `MovieController` already solved this: both of its routes name the parameter `id`, so they merge into one path item, and the inherited `{id:int}` GET is hidden because only one GET can sit there. Performer and studio now do the same. The `{performerForeignId}/works` routes are left alone, since nothing collides with them and renaming would move their generated operation ids without the route changing.
- Every `[RestPostById]` and `[RestPutById]` action was wrong about its status code. `RestController.Created` answers 201 and `Accepted` answers 202, but ApiExplorer only sees the declared return type and documented a 200 for both. An operation filter moves the documented response rather than adding one, which keeps the schema `ActionResult<T>` already inferred. That is 46 corrections: 18 to 201, 26 to 202, and two that already declared the right code alongside a stray 200.
- Request bodies all claimed to be optional. No `[FromBody]` parameter in the codebase is nullable or carries a default, so a request omitting the body is rejected with a 400 before the action runs. ApiExplorer reports them optional because nullable reference types are off in these projects and the compiler cannot tell it otherwise, so a filter supplies it, keyed on the parameter's binding source rather than a list of routes. `POST /login` is untouched, as it takes form fields rather than a body.
- Five endpoints that answer with something other than JSON were documented as answering with nothing. `system/routes` and both log file routes return `text/plain`, the calendar feed returns `text/calendar`, and mediacover returns an image. A return type cannot carry a media type, so each gets `[Produces]` and a `[ProducesResponseType]`. Sonarr annotates its log file action the same way.
- Two endpoints on `ProviderControllerBase` answer with a body the document ignored, across all five provider controllers. `TestAll` builds a `List<ProviderTestAllResult>` and returns it as 200, or as 400 when any provider fails validation; the document declared an empty 200, no 400 at all, and `ProviderTestAllResult` was absent entirely. It is not generic, so the base can name it directly. `RequestAction` returns whatever the invoked action produces, so its schema stays free-form, but saying a JSON body comes back is closer than saying none does.
- `POST /login` and `GET /logout` answer 302 and were documented as 200, a status neither returns. Login also answers 401 when `SignInAsync` throws, carrying `ProblemDetails` because `AuthenticationController` is an `[ApiController]` and nothing suppresses the client error filter. Separately, `POST /api/v3/system/backup/restore/upload` requires a multipart file and declared no body at all, so there was no way to call it from the document. `UploadAndRestore` reads `Request.Form.Files` rather than binding a parameter, so ApiExplorer sees nothing and an operation filter supplies it. The field is documented as `restore`, which is what the UI sends in `RestoreBackupModalContent.tsx`, though the server takes the first file whatever it is called.

Measured against `eros-develop` built alongside:

| | before | after |
|---|---|---|
| document validity | invalid | valid |
| paths | 191 | 186 |
| operations | 274 | 268 |
| empty `200` responses | 64 | 43 |
| request bodies marked required | 0 of 93 | 93 of 94 |
| schemas | 183 | 187 |

The six operations removed are the four frontend GETs and the two duplicate foreign-id templates.
`getPerformerByForeignId` and `getStudioByForeignId` therefore disappear as operation ids, folded
into `getPerformerById` and `getStudioById`, which breaks a client generated from the old document.
Both lookups still work through the merged operation, and the old document did not validate. The five 400s added are the `testall` endpoints, and the two 404s are the performer and studio GETs, which already returned `NotFound()`. `POST /login` and `GET /logout` live in `AuthenticationController` and stay, including the empty security requirement marking them as taking no credentials.

Nothing on the wire changes. Each of the following was checked rather than assumed:

- The route table was dumped from a running instance before and after and is identical for every performer and studio route. The only difference anywhere in the graph is two nodes where `tmdb` and `tpdb` swap ids, which is non-determinism in the numbering.
- A performer was seeded with internal id 7 and foreign id `probe-foreign-id-abc`, then the same four requests run against both builds using the same database file. Byte-for-byte identical responses, and each lookup still resolves as it did.
- `/`, `/movies` and `/system/status` serve index.html, `/login` serves the login page, and `/Content/Images/logo.png` and `/content/Images/logo.png` both answer 200 `image/png`. The UI loads in a browser and navigates with no console errors.
- `[Produces]` is a result filter rather than metadata, so it was driven with deliberately mismatched `Accept` headers. No 406, because these actions return `ContentResult` and `FileResult` and bypass content negotiation. Error paths are unchanged: a missing log file and a missing mediacover both still answer `404 application/problem+json`.
- `TestAll`'s 400 branch was exercised by seeding a download client and then deleting its folders so its test fails. It still returns the full list.
- Forms authentication was enabled with a seeded user to observe the redirects, and the data protection key ring was corrupted to reach the 401.
- A real backup was created and uploaded under two different field names, confirming the server takes the first file regardless of the name.
- Seven further probes across the touched endpoints were run against both builds on the same data directory and returned identical status codes and content types.

One defect turned up that is not fixed here. `POST /api/v3/system/backup/restore/upload` answers 500 with a stack trace when the body is missing, because `Request.Form` throws before the action's own `BadRequestException` guard can run. That is a change in behaviour rather than in the document, so it belongs in its own PR.

<details>

Not a UI change.

</details>

#### Todos

- [x] Tests

`OpenApiFixture` covers the three new filters and the validity fix: it fetches the served document
and asserts the path-parameter rule, that no two paths differ only by parameter name, the 201 and
202 status codes, required JSON bodies, the multipart upload body and the excluded frontend routes.
Dropped onto a build of `eros-develop` all 11 of its assertions fail, each for its own reason.
`Whisparr.Api.Test` passes 172/172 and `Whisparr.Integration.Test` 12/12 for the two OpenAPI
fixtures. The document is also validated by `openapi-spec-validator` and by Redocly, the latter
reporting zero spec-conformance violations where `eros-develop` fails outright.

- Fixes #904

Also related to #882, #887, #893, #898 and #900.


plz12345 (changes requested):
The changes themselves look right. I built the head, served the document and checked it: `openapi-spec-validator` reports OK, and paths (186), operations (268), schemas (187), empty 200s (43) and required bodies (93 of 94) all match your table. Every `[RestPostById]`/`[RestPutById]` action does return `Created`/`Accepted`, so moving the 200 is correct everywhere, and the performer/studio change follows `MovieController`'s existing pattern. `Whisparr.Api.Test` passes 172/172.

**Needed before merge: tests for the new filters.** `SuccessStatusCodeOperationFilter`, `RequiredRequestBodyOperationFilter` and `BackupRestoreUploadOperationFilter` add about 120 lines with no coverage, and the Tests box is ticked. A small fixture that generates the document and asserts the following would cover them, and would also stop this validity fix from quietly breaking again:
- the document validates
- no two paths differ only by parameter name
- PostById/PutById operations document 201/202 and not 200
- `[FromBody]` request bodies are required
- the backup upload has its multipart body

Nits:
- `Fixes whisparr/whisparr-eros#904` can be the bare `Fixes #904`.
- `getPerformerByForeignId`/`getStudioByForeignId` disappear from the document, which breaks clients generated from the old one. That's fine given the document didn't validate before, but it's worth one line in the body.

plz12345 (approved):
Thanks, this covers everything I asked for. `OpenApiFixture` checks each of the requested points against the served document, CI is green including integration_test on all three OSes, and the body now has `Fixes #904` and the operation-id note.

